PR #5, opened 2020-12-27
Using system.Threads and stabilizing update rate.

eslafur:
I don't know where you develop the mod but I included the project files for opening it in visual studio 2019 directly.
only expecting you to install VRC with the MelonLoader dlls in the default path.

besides using a system thread, I opted for measuring how long the update took and wait the approriate amout of time to reach a stable Update rate and not utulize the thread to 100%.
(5-10fps difference on my i7 7700k in a empty world)

my VRC targets a 11ms tickrate, i set the thread to update every 5 ticks being 55ms or ~18hz
the tracking felt a lot less flickery on slower speeds, ~20hz seem like a good middle ground. even blinking still gets recognized.


InconsolableCellist:
5-10 FPS is a big difference. I'd like to try this too

PR #18, opened 2021-04-20
fix lip tracking initialization

uber5001:
for me, lip tracking wouldn't initialize without this change - the mod would simply not mention lip tracking at all


benaclejames:
The lip tracking not initializing is intended behaviour. Until I release the update to support it, the initiazation is commented. You can uncomment the code inside the Initialize method inside SRanipalTrackingInterface from line 52 to line 61 if you want the behaviour to work.

PR #23, opened 2021-04-30
Jaw_Left -> JawLeft to match the rest of the lip shape names

uber5001:


benaclejames:
Fixed in latest release 😄

PR #58, opened 2021-12-02
Addition of Four in One paramaters & TongueSteps

holadivinus:


VRCFaceTracking/Params/Eye/EyeTrackingParams.cs
```diff
@@ -94,16 +94,29 @@ public static class EyeTrackingParams
...
+                if (v2.Left.Openness >= 1 && v2.Left.Widen > 0)
+                    return NormalizeFloat(0, 1, 0.8f, 1, v2.Left.Widen);
+                return v2.Left.Openness < .05 ? -v2.Left.Squeeze : NormalizeFloat(0, 1, 0, 0.8f, v2.Left.Openness);
+            }, "LeftEyeLidExpandedSqeeze", true),
+            new FloatEyeParameter(v2 =>
+            {
+                if (v2.Right.Openness >= 1 && v2.Right.Widen > 0)
+                    return NormalizeFloat(0, 1, 0.8f, 1, v2.Right.Widen);
+                return v2.Right.Openness < .05 ? -v2.Right.Squeeze : NormalizeFloat(0, 1, 0, 0.8f, v2.Right.Openness);
+            }, "RightEyeLidExpandedSqeeze", true),
```
regzo2, line 118: Small typo here! Should be 'Squeeze' for both LeftEyeLidExpandedSqeeze and RightEyeLidExpandedSqeeze.

Also I think you could probably check if the Squeeze value is greater than 0 instead of the Openness value being less than .05 since that might create a dead zone on the blinks (it's probably fine but would rather it be consistent with the other eyelid params).

regzo2 (changes requested):
Few very minor changes. I'm really liking the AI1 params! I think they could be really useful in deprecating some of the combined parameters!

VRCFaceTracking/Params/Eye/EyeTrackingParams.cs
Adjerry91: All the RightEyeLidExpandedSqueeze and LeftEyeLidExpandedSqueeze are with the "u"

holadivinus: fixed 15 days later

PR #71, opened 2022-01-10
Binaryparam infinite scaling + EParams + Eye camera

benaclejames:
Binary parameters are now applicable for all parameters and are infinitely scalable. EParams have been implemented to provide a wrapper for a bool, float and binary parameter for pretty much all parameters.


regzo2:
The only parameters I'm having a small problem with are the binary EyelidExpanded/ EyelidExpandedSqueeze having that (0.8 - 1) on the eyes widen, while blinks get (0 - 0.8). Even on a 4 bit binary param that leaves eyes widen with only 3 steps of movement. I think reverting the binary versions to how they functioned before would be better (binary + 2 bools for widen/squeeze). What do you think?

benaclejames:
> The only parameters I'm having a small problem with are the binary EyelidExpanded/ EyelidExpandedSqueeze having that (0.8 - 1) on the eyes widen, while blinks get (0 - 0.8). Even on a 4 bit binary param that leaves eyes widen with only 3 steps of movement. I think reverting the binary versions to how they functioned before would be better (binary + 2 bools for widen/squeeze). What do you think?

Could this not be achieved using Left/RIghtEyeSqueeze as a separate binaryparam?

regzo2:
Oh it could work with that, but it'd save space to do it the previous way, it's basically a 3-1 combined binary parameter (old) vs separate binary params for eyelid, eyes widen, and squeeze.

regzo2:
I should add it's only the blink/eyes widen part of the EyelidExpanded that's causing a little bit of problem. The squeeze part actually works as intended on binary EyelidExpanded params. Adding an extra binary param for the squeeze would make it take up more space than just using the float variant. Not trying to sound nitpicky 😅. 

Maybe there could just be an EyeLidExpanded like EParam that does the binary logic like the old binary eyelids (returns blink value, then returns eyes widen/squeeze when they are active) that is supplemented with the eyes squeeze/ widen bools

Adjerry91:
Would SqueezeToggle and WidenToggle obsolete with negative param?

regzo2:
Since it's not a combined parameter like the Lip parameters, you would need a separate bool to indicate the different states.

regzo2:
Oh and I also noticed that EyeLidExpanded and EyeLidExpandedSqueeze are pretty much identical EParams

benaclejames:
> Oh and I also noticed that EyeLidExpanded and EyeLidExpandedSqueeze are pretty much identical EParams

Yeah, this was mainly for backwards compatibility, though there's probably a better way to do it

> @benaclejames Would you be opposed if I pushed a change to manually defining the EyeLidExpanded parameters outside of the EParams to revert the binary version?

Go for it!

regzo2:
Not gonna lie, my first push was pretty scary 😂.

I've tested it in-game, and the eyelids are indeed functioning as expected now! Please check over it and make sure I didn't make any goofs.

regzo2:
Accidentally named CombinedEyeLidExpanded as LeftEyeLidExpanded, should be fixed now 😅

benaclejames:
Sweet! Still got a couple questions though 😄 

https://github.com/benaclejames/VRCFaceTracking/blob/d92bfc35009f6e3331f88eb6ae0fd4394159f34e/VRCFaceTracking/Params/Eye/EyeTrackingParams.cs#L61
https://github.com/benaclejames/VRCFaceTracking/blob/d92bfc35009f6e3331f88eb6ae0fd4394159f34e/VRCFaceTracking/Params/Eye/EyeTrackingParams.cs#L71

Why do both these lines reference data for the left eye when they're for the right and combined eye?

Also, do you think it'd be possible to have just kept the EParams but changed the logic inside them to use the same as the old binary params? Seems we're now defining multiples which may cause issues

regzo2:
Whoops, they're mean't to use their respective data reference. Was wondering why I couldn't control them that well 😅. @Adjerry91 pointed that out too.

Also if we use the binary logic on floats, in combo with the bool toggles, I think the interp over the network will mess them.

I think defining a separate Eparam with the logic would be the best option if you want to set then up as an EParam, either redefining one of the eyelids or new param (idk what would be best at this point)

regzo2:
All righty fixed those two errors, forgot to push those when I did them!

Adjerry91:
I was wondering why my eyelid were acting funny

benaclejames:
@regzo2 Just pushed some final cleanup using EParams, combined with your blending logic between widen, openness and squeeze, If it all looks good, I think we're ready to accept this PR. Thanks
https://github.com/benaclejames/VRCFaceTracking/commit/827fea8f9d3999161a251f5f2cd4a151b9295e86

regzo2:
@benaclejames Existing avatars using the float params of EyeLidExpanded/EyeLidExpandedSqueeze will have to be redone if you want to use the binary logic between all the variants of the parameters. I think it will work otherwise for the binary/bool version of the params. Everything else looks good!

regzo2:
Maybe it would be best to make a separate EParam for the old binary logic, and a separate one for how it worked on the float versions.

benaclejames:
Hmm, you're right. There's gotta be a better way to go about this short of redefining a load of parameters

Adjerry91 (commented):
This code does not work with EParams as the math is completely different for floats and binary system.

regzo2:
@benaclejames Noticed you updated the EParam constructor, so I went ahead and setup the eyelid params to use the new constructor. Let me know what you think!

benaclejames:
Well, I'm happy with the eye image stuff. If you're happy with the binary params, we'll go ahead and complete this!

Adjerry91:
VRChat_1920x1080_2022-01-29_20-28-07 144
I do not think this working correctly on v1.3.1.1

regzo2 (approved):
Everything looks good now!

Adjerry91:
Working for me also!

Adjerry91:
crap found issues with eye expanded code

Adjerry91:
pushed it separately from the merge to remove squeeze component of EyeExpanded bools

PR #72, opened 2022-01-13
add some computed lip shapes

ftfft:


Adjerry91:
Is this same as the combined left/right params on #73?

ftfft:
> Is this same as the combined left/right params on #73?

yep basically, just some small details in implementation that are different.

PR #73, opened 2022-01-18
Implement combined left/right params

AeroScripts:
Many people seem to be using only 1 side of the combined params to power both sides, which isnt ideal for some params. This PR creates new combined params that average the left and right side for these, which should give more consistent resullts across different users. Its possible taking the biggest param (math.max) would be better than taking the avereage for some of these

There is some similar code implemented in #72 for combining/averaging shapes also.

### DIFF
--- a/VRCFaceTracking/Params/Lip/LipShapeConversion.cs
+++ b/VRCFaceTracking/Params/Lip/LipShapeConversion.cs
@@ -3,7 +3,13 @@ using ViveSR.anipal.Lip;
 
 namespace VRCFaceTracking.Params.Lip
 {
-    public class PositiveNegativeShape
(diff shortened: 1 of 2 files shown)


Adjerry91:
I do not see any problems with the code with the binary params update

Adjerry91:
Tested code and does work. It actually seams more stable values with both of them combined.

benaclejames (approved):
LGTM!

Adjerry91:
Merged commit went to the binaryparam-infinite-scaling branch and now is gone from the main repo.

Adjerry91:
I am coping the code from changes made and going to make separate commit on main

PR #84, opened 2022-04-17
Adding error message when loading an external module fails

InconsolableCellist:
Was quite handy to find out why VRCFTVarjoModule wasn't working (the recent Interface changes)


benaclejames:
Sweet! Would you mind moving this check and message to the Assembly.LoadFrom on line 89?

PR #88, opened 2022-04-26
Re-Enable and Fix Pimax Eye Tracking

0x8080:


benaclejames:
Noticed you were using UnityEngine here, would it be possible to do this without relying on Unity libs?

0x8080:
> Noticed you were using UnityEngine here, would it be possible to do this without relying on Unity libs?

Yeah, I can use Math.atan() to perform the arc-tangent operation needed.

PR #90, opened 2022-05-29
Fixed a crash when VRCFT is used together with VOR

m3gagluk:
Apparently a simple Directory,Exists() call always returns true even when a target directory doesn't exist (at least that happens in the AppData special folders). I had to wrap the code into a try...catch block as a workaround for VOR creating its own directory in the VRC's OSC directory.


Nekomiii (changes requested):
with my friend i have try to change with the fixe that you build and it didn't work the app "vrcft" keep crashing with vor activated and the eye tracking can't be enable with the fixe, than lips tracking can be enable correctly !

benaclejames:
Unable to reproduce the bug you're trying to fix. Directory.Exists appears to work fine for any directory I throw at it. Feel free to re-pr this if you can give an example of when the Directory.Exists check fails.

PR #91, opened 2022-06-30
Reduce memory consumption when working with large camera images

m3gagluk:
I found an issue regarding higher memory consumption with a lot of unncessary GC activity while using the cutting edge build of VRCFT with a Varjo module that supplies camera data to the app. The issue is most noticeable when a module provides a high-resolution image - I know that VRCFT doesn't need such an high-res image but the issue still persits


benaclejames (approved):
Awesome! Thank you!

PR #95, opened 2022-07-29
Adding external modules to README

InconsolableCellist:
Because I was having trouble finding it


regzo2:
Here are all of the modules that I know of that exist, might as well add these while we're at it:
Varjo (VRCFT Varjo Module)
HP Reverb G2 Omnicept (Eye Only. VRCFT Omnicept Module)
Pimax Droolon Pi1 (VRCFT Pimax ET Module)
Desktop Apple AR (using iPhone X / iPad Pro 3rd Gen or newer) (VRCFT LiveLink Module)
Desktop Webcam (NoVRCFT (Module))
Desktop Kinect (VRCKinectFaceTracking (Module))

PR #96, opened 2022-08-03
Chillout VR Support with OSC Mod

bizzclaw:
This will check to see if CVR is running and load the avatar config file from that instead of the VRChat. It's programemd to work off of the CVROSC mod by KaleidonKep99 here

https://github.com/KaleidonKep99/CVROSC


bizzclaw:
This is obsolete thanks to Tiger's CVRParamlib which just puts things directly in the VRChat folder.
VRCOSC also got deprecated anyways, so this isn't needed.

PR #100, opened 2022-08-11
Expose pupil diameter from 0 to 1 cm

hai-vr:
- Expose pupil diameter as EyesPupilDiameter to provide tighter control for avatar owners. The existing parameter, EyesDilation, relies on historical eye diameter data acquired since the last avatar change to determine the minimum and maximum eye dilation, which can vary depending on the recent user activity, calibration error, and lighting conditions.
  This makes it difficult to create an eye representation using EyesDilation that best reflect physiological effects that the user may be experiencing during a specific session.
- Due to a lack of available information regarding the iris size on the Vive Pro Eye, the currently exposed diameter is in an absolute unit rather than relative. Exposing other parameters leveraging on other APIs such as the Varjo Aero is out of scope here.
- A pupil diameter of 1 cm is represented with the value of 1.
  According to various sources, human pupil size may contract and expand, reaching diameters between 2mm and 9mm (with no significant digits specified).
  For convenience, the maximum value of 1 represents 1 cm.

### DIFF
--- a/VRCFaceTracking/Params/Eye/EyeTrackingParams.cs
+++ b/VRCFaceTracking/Params/Eye/EyeTrackingParams.cs
@@ -33,6 +33,7 @@ namespace VRCFaceTracking.Params.Eye
             #region Dilation
             
             new EParam(v2 => v2.EyesDilation, "EyesDilation"),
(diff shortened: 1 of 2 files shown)


BenjaminZehowlt:
Could you please rebase your branch the latest master commit.

PR #102, opened 2022-09-10
Start automatically removing 'downloaded from the internet' identifiers from custom modules

m3gagluk:
That PR is going to fix the main issue of many first time users of custom VRCFT modules that is forgetting to unblock the downloaded module. Yay


Chickenbreadlp:
I'd suggest that if deleting the Zone File fails, that we shouldn't try to load the DLL and instead pretend as if it wasn't even there. It should prevent the app from crashing and would open up the possibility for a proper Error Message to appear

m3gagluk:
Sorry, didn't see the new comment notification. 'Broken' modules will be simply skipped from now

PR #104, opened 2022-10-19
Fix for VRCFaceTracking crash with port binding

KaleidonKep99:
Right now VRCFaceTracking fails to bind the input port if it's already bound to another app, preventing it from working. Setting it to reuse the address will allow VRCFaceTracking to boot even when another app already bound itself to the port.


PinkP4nther:
This is a security risk and has indeterminate behaviour >.<

KaleidonKep99:
> This is a security risk and has indeterminate behaviour >.<

What would be the security risk here?
All VRCFaceTracking does is send booleans and floats and receive strings to and from VRChat respectively.

benaclejames:
Rejecting due to OSCQuery implementation coming soon

PR #109, opened 2022-11-15
Top-down Refactor of VRCFaceTracking to work more as an interface.

regzo2:
This draft is based on the VRCFT Proposal listed here: https://docs.google.com/spreadsheets/d/118jo960co3Mgw8eREFVBsaJ7z0GtKNr52IB4Bz99VTA

The rest of this draft will be covering what I best believe the most ideal way forward for VRCFaceTracking as an interfacial software for hardware/software developers and VRChat. A summary of this draft is near the bottom of this page.

### VRCFaceTracking's Goal as an Interface

As mentioned above, I think we should approach development of VRCFaceTracking as an interface for facial tracking in VRChat. The main three goals of VRCFaceTracking have been so far as: standardize facial expressions (AKA blendshapes) and parameters (with SRanipal as the standardization basis) and compress and standardize the parameters to be useable in VRChat for VRCFaceTracking supported avatars. There have been a few setbacks with this approach thus far, and those will be outlined as we move through this draft.

As VRCFaceTracking has been developing, the main limiting factor so far is being limited to SRanipal's facial expression standard. While for the most part this has worked very well for VRChat and our short-term goals of getting face tracking working in VRChat, newer devices that are able to provide much more intricate and detailed facial tracking are problematic: the main points of contention being: having not enough base parameters to interface with and how the combined parameters have been developed to interface with SRanipal's unique quirks in mind.

I believe that we can use Unified Expressions as a launchpad to make and implement a face tracking standard that is not only robust enough to capture the fidelity of all hardware and software face tracking solutions, but also to combine all of the popular facial tracking standards used today into a ubiquitous standard and have the potential to build upon them if the need arises. This allows us to have backwards compatibility with many popular face tracking standards, as well as forward compatibility with any future revisions with the VRCFT Proposal parameters. We also would like to make far better documentation and more discretely available resources alongside the new standard/s. Module developers will be able to be able to fully implement previously unavailable facial expressions within VRCFaceTracking's module system.

Combined parameters may need to be updated as well to remove some non-recommended configurations, and to add in support for the new parameters. These should be keenly observed when testing with different hardware configurations and fidelities.

**I also think that the sooner we integrate with the VRCFT Proposal parameters, the easier and speedier it will be to develop both references/documentation and other issues that may arise as well.**

### What VRCFaceTracking can improve upon

I also think this will be a really prime time to incorporate a calibration / normalization tool or feature into VRCFT. I have a WIP calibration tool in VRCFT for calculating parameter multipliers, but I believe that an additional measure would be to have dynamic adjust-fitted normalization that runs continuously in the software as well. We should ideally not let users have direct control of how the parameters get controlled or normalized (as that would cause inconsistent user-side setups to arise), but having VRCFaceTracking do calibration automatically or semi-automatically will be healthy for both end-users and module development. Module developers by that change would expectedly only need to feed the correctly mapped data and leave the normalization and parameter driving to VRCFaceTracking. The normalization and/or calibration could also help keep parameters far more consistent than having module developers or internal developers set and mandate parameter multipliers and normalization by themselves.

### The Goal of the VRCFT Proposal Parameters

To expand upon what the aforementioned VRCFT Proposal is, it is as mentioned a unification of 3 popular face tracking standards: SRanipal, ARKit/AR52/Perfect Sync, and Movement by Meta. For nearly all of the shapes we prefer to make the blendshapes match up identically to the standards by matching up the parameter's up with each other. The shapes we would prefer to use ARKit and Movement blendshapes over SRanipal (with some small exceptions) due to the former standards' having much more intricate facial expressions, but for most of the shapes they end up being the same. The parameter naming convention is carried over from SRanipal for most shapes, but this doesn't affect the compatibility of the other 2 standards, and their ability to interface into the proposal.

Currently we do not have a blendshape reference. As the proposal parameters get finalized, a humanoid head mesh will be created to work with the VRCFT Proposal parameters, which should help implement these parameters. The proposal also falls in-line with popular standards for the most part, so many existing assets can and should be used as a testbed as well.

### Conclusion

To summarize: I believe the VRCFT Proposal and the Unified Expressions implementation will be extremely important to capitalize on these things: making base parameters well defined (have discrete documentation and references), consistent (with normalization compensation within the software), and mutually exclusive (this includes base parameters being well defined and isolated, and this goes for the current combined parameters as well) and having the combined parameters and parameter setups be ubiquitously supported across any hardware configuration using built-in calibration and/or normalization.

Thank you for reading, please consider this as we implement the new Unified Expressions system into VRCFaceTracking.


regzo2:
Hmm I'm starting to think we should have the eye and lip shapes just be accessed in one UnifiedExpression, instead of how it currently is with the Eyes and Lips seemingly having their own structures. I have the Eye gaze expressions commented out as I'd expect there to be a more appropriate data type to accept those, but I say it's fair game with the rest of the expressions being unified into one structure.

benaclejames:
My only request is backwards compatibility 😄

regzo2:
Backwards compatability for avatars and existing modules is something we should definitely retain, especially for keeping existing stuff working without the need to update. I'll think about a best way to approach this tomorrow lol😥

Adjerry91:
Aliases dictionary will cover backwards compatibility linking. Most of them be one to one.

regzo2:
Looking through the code on the Quest Pro module, the ALVR module doesn't actually use the Update function on LipTrackingData or EyeTrackingData, and instead it just directly modifies the LipShape and Eye structures respectively. Is this something we should expect modules to do? I was thinking of using the Update functions to incorporate calibration in VRCFT, but the ALVR module would effectively work around it in it's current implementation.

regzo2:
I did a lot of work refactoring a lot of stuff especially in UnifiedTrackingData and the parameters system in regards to UnifiedExpressions in VRCFT today. I am still working on it but I will have a very detailed write-up of all the changes I've done thus far with a pretty huge commit to my fork (I want to get it working before committing for reasons 😥), but it touches a lot of VRCFT so I'm really hoping I'm going in a positive direction with the changes.

I have tried putting a lot of thought into how I was implementing these changes and making it more clear and accessible for both module developers and for parameter improvements and changes. A lot of the changes were pretty much cutting away SRanipal dependencies in the base VRCFT functions, so a lot of the functions now do not use SRanipal data structures (yay). 

In place of those removed structures I have implemented additional custom structures to fill their spots, and they are as follows: UnifiedExpressions (effectively a LipShape replacement, mostly implemented), UnifiedExpressionsData (LipData replacement), UnifiedEyeData (replaces EyeData), UnifiedSingleEyeData (is effectively SingleEyeData). I currently have it so that module developers can use these structures (without having SRanipal dependencies) to use the update functions of EyeTrackingData and ExpressionsTrackingData (previously LipTrackingData, and also to add that the SRanipal module was the only one effectively using the update function after looking a little deeper), they can do this in a few ways, such as creating UnifiedExpressionsData and UnifiedEyeData objects and feeding them into the Update functions in UnifiedTracking. I believe this will be the ideal mode of updating parameters in VRCFT, I'm still not sure if we should still keep the ability to touch the LatestShapes and EyeData directly, but currently I still have them accessible as there isn't currently a reason to have them locked off from use in modules (maybe if there is a calibration or normalization step in here eventually, but I would like to know if this would be a positive change.). Let me know if this approach seems like a good idea! I'm still a bit hairy on this implementation, and I believe there may be a better way to approach how module devs interface with UnifiedTracking in my fork, but this seems like it would work.

I've also effectively separated out the Unified Expressions parameters system from the Lip and Eye parameters for the time being (in the current main `Quest-Pro` branch UnifiedExpressions is only designed as a LipShape replacement/enhancement, whereas in my current work it's a separate system from LipShape), and I want to eventually drive the Unified Expressions into the old parameters. Currently modules have to drive two separate 'LatestShapes' (one for Unified Expression and one for SRanipal) for full backwards compatibility, but eventually I want to have Unified Expressions drive SRanipal with a compatibility layer to avoid that. I think I have a pretty solid approach for this so far, at least in regards to the UnifiedExpressions working by itself.

Currently my priority is getting this version of Unified Expressions to work by itself, in isolation, and then integrate backwards compatibility afterwards (which I'm anticipating will not be too much hassle thankfully), and lastly test and create some combined parameter combos for Unified Expressions.

Also if you're worried: this isn't like how I first yeeted Binary params into the code, I have learned a lot since then 😴. Okay hopefully this wasn't too rambly, I just wanted to let you guys know what I am working on.

regzo2:
Okay so with a little bit of sleep I have some thoughts on how maybe modules should interface with the LatestShapes/Eye with what I did, I just realized that the system I had put in would mean that modules would have to send the full data of UnifiedEye instead of just directly interfacing with the structure itself already. I can already see an issue being that multiple eye modules would override each other by sending their own datas on update (which is really bad for keeping flexibility in modules). 

What I think I will do is have LatestShapes and the eye structs on EyeTrackingData be directly accessible by modules, and then use the Update function to tell VRCFT to update a more internal/inaccessible structure using the passed data. This means that modules can send partial data and just tell VRCFT to update accordingly, so now modules would function similar to before but pretty much with an update button to let VRCFT know that the data can be merged more internally.

regzo2:
I've actually thought of another possible thing modules could be able to do: request 'ownership' of tracking parameters! I think access to the eye structure will still be handled like it is now (as in modules ask for ownership of the entire Eye structure now that it is more simpler in my fork's implementation), but with the rest of Unified Expressions, modules could ask for specific parameters to 'own' (such as in the case that one module has an incomplete interface, and another one fills in the remaining expressions). Would be really cool I think, but might be a bit of a pain with the current implementation to do that. I'll focus on getting this fork working first before even considering doing that

regzo2:
Okay pretty happy with the state of the interface now, gonna try playing with some normalization stuff

regzo2:
Created a Mutator class for implementing calibration, normalization, smoothing, and other parameter meddling algos. I'm not entirely sure of the best approach for this class but I'd like it so that modules could possibly override some of the methods used in the class (and I may make it abstract so that can be possible.)

regzo2:
Renamed the PR to better reflect the actual developmental scope of my fork.

regzo2:
Moved into review! Please let me know if there are any issues with using this build, or if there is anything that needs to be touched over (I'm not perfect so I hope any changes I have made are either intuitive or have been for the better). 

Everything previously integrated in VRCFaceTracking should be usable in the newer builds. I will have a build with all of the modules on my fork built for the new module and parameter system. Currently only have the VRCFT 'SRanipal' standard accessible at the moment, and will be working on having the Unified Expressions base parameters be accessible very soon.

regzo2:
Here is everything this fork does at a glance so far, and please let me know if there are any systems that I've touched that need to be reconsidered:

Unified Tracking system:

- Overhauled VRCFT to use the VRCFT Proposal parameters (dubbed Unified Expressions) from the document in the original post, with some minor adjustments to the standard after several tests with meshing the existing standards into it. Unified Expressions is a community driven parameter system and standard that combines the most popular face tracking standards into one.
  - Existing modules will be incompatible with this new parameter system, and will have to be updated to drive the newly updated parameters. I really hope that the changes brought with the new system are worth breaking backwards compatibility with legacy versioned modules.
- Entirely independent parameter system that has no dependencies on existing tracking systems (SRanipal), and emulates into the currently implemented parameter conversion system in VRCFT. There is no noticeable performance difference between the old system and the new emulation into the old system. The new parameter system and conversions should perform on par with the existing implementation. (They were already very performant so there are no complaints from me!). Currently there isn't any way to access these in the current iteration as their output isn't completely finalized. All VRCFT SRanipal parameters are emulated so all existing VRCFT avatars are still usable.
- New Mutator system that acts on the buffer between a tracking interface and the VRCFaceTracking parameter conversion system. This will allow VRCFT to implement a myriad of parameter mutations and corrections, such as calibration, smoothing, normalization, or any other parameter meddling you can think of! VRCFaceTracking automatically calibrates normalization on the fly, and will be updated to also have a calibration period with a fine-tuning algorithm afterward. I'm still not sure of the best way to implement this system, but Module developers can currently adjust values inside the Expression parameters to tweak the calibration and smoothing values if they want it to line up with their tracking system perfectly (though I would still prefer to have VRCFT handle calibration values).
- Did some huge reorganizations of UnifiedTrackingData (now globally contained under UnifiedTracking). UnifiedTracking is now a lot more tidied up to indicate it's bridging nature (along side LibManager) for parameters from tracking interfaces to VRCFT's parameter conversion and mutations.
- Modules can choose when to push parameters into VRCFT's parameter conversion using an UpdateData() call.

Module system:

- Overhauled the library manager so that it will load all module paths and let you set what module assemblies get loaded in a set order of your choosing. The module load order will be saved into a config file in the future with a config file to have automatic startup with end-user selected modules.
- A lot of small tweaks to let modules overlap each other elegantly.
- Some small adjustments to the ExtTrackingModule abstract to complement the change. 
- VRCFaceTracking can now be portable! Including a CustomLibs directory next to the EXE will cause VRCFaceTracking to load modules from the portable folder as well, and prioritizes the portable modules over installed (AppData) ones.

Misc:
- SRanipalExtTrackingInterface is now it's own separate module. After cutting out SRanipal dependencies in VRCFaceTracking I was able to move this into it's own separate class library.
- The UI was updated to let some new features be usable, but this iteration of the UI is intended to be replaced by a much bigger overhaul in the near future.

VRCFaceTracking/Params/Expressions/UnifiedExpressionsMerger.cs
```diff
@@ -0,0 +1,198 @@
...
+            new EParam(exp => Math.Min(1, exp.Shapes[(int)UnifiedExpressions.BrowInnerUpLeft].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.BrowOuterUpLeft].AdjustedWeight) -
+                Math.Min(1, exp.Shapes[(int)UnifiedExpressions.BrowInnerDownLeft].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.BrowOuterDownLeft].AdjustedWeight),
+                "v2/LeftBrowY"),
+
+            new EParam(exp =>
+                (Math.Min(1, exp.Shapes[(int)UnifiedExpressions.BrowInnerUpRight].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.BrowOuterUpRight].AdjustedWeight) -
+                Math.Min(1, exp.Shapes[(int)UnifiedExpressions.BrowInnerDownRight].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.BrowOuterDownRight].AdjustedWeight)) +
+                (Math.Min(1, exp.Shapes[(int)UnifiedExpressions.BrowInnerUpLeft].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.BrowOuterUpLeft].AdjustedWeight) -
+                Math.Min(1, exp.Shapes[(int)UnifiedExpressions.BrowInnerDownLeft].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.BrowOuterDownLeft].AdjustedWeight)) / 2.0f,
+                "v2/LeftBrowY"),
```
Adjerry91, line 79: v2/BrowsY

VRCFaceTracking/Params/Expressions/UnifiedExpressionsMerger.cs
```diff
@@ -0,0 +1,198 @@
...
+using System.Threading.Tasks;
+
+namespace VRCFaceTracking.Params
+{
+    public static class UnifiedExpressionsMerger
+    {
+        public static readonly IParameter[] UnifiedCombinedShapes =
+        {    
+            // Unified Eye Definitions
+
```
Adjerry91, line 14: May need to name v2 to something else that more useful to understand in the parameter drop down box. This should be permanent goal to have it in a menu for organization. So  i.e. FT/LeftEyeX

VRCFaceTracking/Params/Expressions/UnifiedExpressions.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        #endregion
+
+        #region Nose Expressions
+
+        NoseDilationRight, // Right side nose's canal opens up.
+        NoseDilationLeft, // Left side nose's canal opens up.
+
+        NoseWrinkleRight, // Outer part of the right side nose pulls upward.
+        NoseWrinkleLeft, // Outer part of the right side nose pulls upward.
```
Adjerry91, line 60: NoseSneerLeft and NoseSneerRight

VRCFaceTracking/Params/Expressions/Legacy/Lip/UnifiedSRanMapper.cs
```diff
@@ -0,0 +1,66 @@
...
+            { SRanipal_LipShape_v2.MouthLowerDownLeft, exp => exp.Shapes[(int)UnifiedExpressions.MouthLowerDownLeft].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.LipFunnelLowerLeft].AdjustedWeight },
+
+            { SRanipal_LipShape_v2.MouthUpperInside, exp => (exp.Shapes[(int)UnifiedExpressions.LipSuckUpperLeft].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.LipSuckUpperRight].AdjustedWeight) / 2.0f },
+            { SRanipal_LipShape_v2.MouthLowerInside, exp => (exp.Shapes[(int)UnifiedExpressions.LipSuckLowerLeft].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.LipSuckLowerRight].AdjustedWeight) / 2.0f },
+
+            { SRanipal_LipShape_v2.MouthLowerOverlay, exp => exp.Shapes[(int)UnifiedExpressions.MouthRaiserLower].AdjustedWeight},
+
+            { SRanipal_LipShape_v2.TongueLongStep1, exp => exp.Shapes[(int)UnifiedExpressions.TongueOut].AdjustedWeight},
+            { SRanipal_LipShape_v2.TongueLongStep2, exp => exp.Shapes[(int)UnifiedExpressions.TongueOut].AdjustedWeight},
```
Adjerry91, line 46: TongueOut would be TongueLongStep1 and TongueLongStep2 both 1. So TongueLongStep1 should be half of Tongue Out and TongueLongStep2 should also be half of TongueOut

VRCFaceTracking/Params/Expressions/Legacy/Lip/UnifiedSRanMapper.cs
```diff
@@ -0,0 +1,66 @@
...
+            { SRanipal_LipShape_v2.CheekPuffRight, exp => exp.Shapes[(int)UnifiedExpressions.CheekPuffRight].AdjustedWeight },
+            { SRanipal_LipShape_v2.CheekSuck, exp => (exp.Shapes[(int)UnifiedExpressions.CheekSuckLeft].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.CheekSuckRight].AdjustedWeight) / 2.0f },
+
+            { SRanipal_LipShape_v2.MouthUpperUpRight, exp => exp.Shapes[(int)UnifiedExpressions.MouthUpperUpRight].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.LipFunnelUpperRight].AdjustedWeight },
+            { SRanipal_LipShape_v2.MouthUpperUpLeft, exp => exp.Shapes[(int)UnifiedExpressions.MouthUpperUpLeft].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.LipFunnelUpperLeft].AdjustedWeight },
+            { SRanipal_LipShape_v2.MouthLowerDownRight, exp => exp.Shapes[(int)UnifiedExpressions.MouthLowerDownRight].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.LipFunnelLowerRight].AdjustedWeight },
+            { SRanipal_LipShape_v2.MouthLowerDownLeft, exp => exp.Shapes[(int)UnifiedExpressions.MouthLowerDownLeft].AdjustedWeight + exp.Shapes[(int)UnifiedExpressions.LipFunnelLowerLeft].AdjustedWeight },
```
Adjerry91, line 38: What is the basis for using LipFunneler with MouthLowerDown and MouthUpperUp shapes?

VRCFaceTracking/Params/Expressions/Legacy/Lip/UnifiedSRanMapper.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace VRCFaceTracking.Params.Lip
+{
+    class UnifiedSRanMapper
+    {
+        private static Dictionary<SRanipal_LipShape_v2, Func<UnifiedTrackingData, float>> ShapeMap = new Dictionary<SRanipal_LipShape_v2, Func<UnifiedTrackingData, float>>
+        {
+            { SRanipal_LipShape_v2.JawRight, exp => exp.Shapes[(int)UnifiedExpressions.JawRight].AdjustedWeight },
+            { SRanipal_LipShape_v2.JawLeft, exp => exp.Shapes[(int)UnifiedExpressions.JawLeft].AdjustedWeight },
+            { SRanipal_LipShape_v2.JawForward, exp => exp.Shapes[(int)UnifiedExpressions.JawForward].AdjustedWeight },
+            { SRanipal_LipShape_v2.JawOpen, exp => exp.Shapes[(int)UnifiedExpressions.JawOpen].AdjustedWeight - exp.Shapes[(int)UnifiedExpressions.MouthClosed].AdjustedWeight },
```
Adjerry91, line 13: Most likely will need to look at JawOpen and MouthApeShape in testing as this can be tricky to make work as good as native SRanipal mouthApeShape

VRCFaceTracking/Params/Expressions/Legacy/Eye/EyeTrackingParams.cs
```diff
@@ -0,0 +1,148 @@
...
+                ? exp.Shapes[(int)UnifiedExpressions.EyeWideLeft].Weight > 0
+                : exp.Shapes[(int)UnifiedExpressions.EyeWideRight].Weight > 0
+                , "EyesWidenToggle"),
+
+            new BoolParameter(exp => exp.Shapes[(int)UnifiedExpressions.EyeSquintLeft].Weight > 0, "RightEyeSqueezeToggle"),
+            new BoolParameter(exp => exp.Shapes[(int)UnifiedExpressions.EyeSquintRight].Weight > 0, "RightEyeSqueezeToggle"),
+            new BoolParameter(exp => exp.Shapes[(int)UnifiedExpressions.EyeSquintLeft].Weight > exp.Shapes[(int)UnifiedExpressions.EyeSquintRight].Weight
+                ? exp.Shapes[(int)UnifiedExpressions.EyeSquintLeft].Weight > 0
+                : exp.Shapes[(int)UnifiedExpressions.EyeSquintRight].Weight > 0
+                , "EyesSquintToggle"),
```
Adjerry91, line 139: Squint and squeeze are not the same thing when tracked. This should not be included with the old eye tracking bool switch

VRCFaceTracking/Params/Expressions/UnifiedExpressions.cs
```diff
@@ -0,0 +1,201 @@
...
+        #region Eye Expressions
+
+        // 'Biometrically' accurate data that is included with UnifiedEye
+        //EyeClosedRight, // Closes the right eyelid
+        //EyeClosedLeft, // Closes the left eyelid
+        //EyeDilationRight, // Dilates the right eye's pupil
+        //EyeDilationLeft, // Dilates the left eye's pupil
+        //EyeConstrictRight, // Constricts the right eye's pupil
+        //EyeConstrictLeft, // Constricts the left eye's pupil
+
```
Adjerry91, line 33: Add Squeeze. Squint and squeeze are not the same

VRCFaceTracking/MainStandalone.cs
```diff
@@ -90,15 +101,24 @@ public static void Initialize()
...
-                Logger.Msg("Config file parsed successfully! " + _relevantParamsCount + " parameters loaded");
+
+                // Reset calibration on parameter data.
+                UnifiedTracking.Data.ResetCalibration();
+
+                _relevantParamsCount = _relevantParams.Count() + _relevantParams_v2.Count();
+                Logger.Msg("Config file parsed successfully! " + _relevantParamsCount + " parameters loaded.");
+                
+                if (_relevantParams.Count() > 0)
+                    Logger.Warning(_relevantParams.Count() + " legacy parameters loaded. These are undocumented and outdated parameters.");
```
Adjerry91, line 121: I am assuming there is going to be switch somewhere here as looks like this would always say legacy parameters are loaded.

regzo2: All _relevantParams are the original SRanipal parameters (pulled from UnifiedTracking.AllParameters_V1), while _relevantParams_V2 are all pulled from V2. If _relevantParams happens to contain any parameters we would know that the original parameters were loaded in some capacity.

VRCFaceTracking/Params/Expressions/Legacy/Lip/UnifiedSRanMapper.cs
regzo2: That's right, let me make that change. It should still work consistently when using TongueOut from the SRanipal combined parameters. I'll make it so that TongueStep1 uses 0-0.5 range and Step2 uses 0.5-1 range from Unified TongueOut

VRCFaceTracking/Params/Expressions/UnifiedExpressions.cs
regzo2: All right sounds good! The initial name is from FAQS and this change would make it line up closer with the ARKit naming scheme.

VRCFaceTracking/Params/Expressions/UnifiedExpressionsMerger.cs
regzo2: All right will make that change, thanks for catching!

VRCFaceTracking/Params/Expressions/UnifiedExpressionsMerger.cs
regzo2: v2/... is to indicate it is using the new standard, I can make it so that you would be able to attach any extensions you would want (in the case of organization).

VRCFaceTracking/Params/Expressions/Legacy/Lip/UnifiedSRanMapper.cs
regzo2: Lip Funneler opens the upper and lower lips, which Overturn does not have baked in. This will force the tracking interface to behave as SRanipal does with the respective SRanipal shape. The SRanipal module does the inverse of this operation and it seems to work as intended. I have yet to do a A/B test between old and new for this iteration but mathematically it should be able to be isolated properly.

VRCFaceTracking/Params/Expressions/Legacy/Lip/UnifiedSRanMapper.cs
regzo2: It seems to work as intended from my own testing. It should be mathematically consistent as far as removing the Ape part directly from SRanipal as MouthClose is using the same value as MouthApe from SRanipal. This also forces any new tracking interfaces to behave like SRanipal would when using the old parameters. Sending the data to JawOpen just happens to have both Sranipal JawOpen and SRanipal MouthApe baked into the value as I'm considering them to both open the jaw from the SRanipal tracking interface.

We could consider making a new output parameter named MouthApeShape for v2 that acts like SRanipal Ape but I think we should try to avoid losing fidelity in the new standard's combined params.

VRCFaceTracking/Params/Expressions/Legacy/Eye/EyeTrackingParams.cs
regzo2: >Squint should not be included with the old eye tracking bool switch

Oops looks like I accidentally renamed the output of this param. 

> Squint and squeeze are not the same thing when tracked.

As for Squint and squeeze not being the same I have elaborated on that in another comment thread.

VRCFaceTracking/Params/Expressions/UnifiedExpressions.cs
regzo2: Could you elaborate further, I'm not entirely sure what you consider them different. Here is what I've considered when using this parameter for this particular v1 expression:

Unified Squint = Quest Pro Lid Tightener = AR52 Eye Squint = (SRan Eye Squeeze + SRan Eyes Closed)

SRanipal EyeSqueeze from what I can tell from using and from visual references is a combination of squinting the lower eyelid muscles (Ala Quest Pro Lid Tightener) and closing the eyes (Ala Quest Pro EyeBlink). The only issue is that SRanipal EyeSqueeze only activates when the eyes are closed and acts as it's own threshold beyond closed eyes. I think even from our own SRanipal documentation that this would be the case as well https://github.com/benaclejames/VRCFaceTracking/wiki/Blend-Shapes-Setup#sranipal-eye-shapes

When messing around with the High Fidelity Quest blendshapes reference it does seem to indicate that when Lid tightener and Eyes closed are active at the same time that there would be a corrective shape that makes them visually identical or of similar function to SRanipal EyeSqueeze.

https://user-images.githubusercontent.com/74634856/214394165-bbd35bcf-e672-4026-a535-d1c82d50190c.mp4

I think the only problem is that this SRanipal Squeeze parameter kind of is locked in to behave properly for SRanipal and doesn't take into account having Eye Blink and Squint be mutually exclusive (which is true for Quest Pro and every AR52 implementation). The SRanipal combined eye parameters would have to be fundamentally changed in order for it to behave as expected with the new standard.

SRanipal Squeeze does track differently but should fundamentally control to the same or very similar shapes to our standard. I am not entirely sure if adding an extra parameter that only SRanipal could solely take advantage of would be necessary if it already has such a huge overlap with the existing parameters. I think we should be building Unified to be able to control SRanipal parameters in a high fidelity way rather than allowing SRanipal to just pass-through directly.

Let me know if you think I am wrong about these assumptions.

regzo2: I also think that the SRanipal emulator could and should be able to replicate SRanipal Eye Squeeze's behavior, I just didn't have a tracking interface that let me play around with the emulator beyond the SRanipal interface until I made the MeowFace interface for testing it.

regzo2: Okay after playing around with the Unified data a bit I believe I have a solution that can solve these two issues, which will allow any tracking interface to still work really well even in the emulated combined eyelids.

I will treat the Squeeze section and Blink + Widen sections of the parameters in their own distinctive ways: 

Blink + Widen will still have Squint weighted into the openness (as it does affect the SRanipal eye shape slightly on all interfaces, it's just that SRanipal doesn't squint with eyeSqueeze) but to a much lesser extent than it currently is enforced in. When the eyes are closed then we can enforce Squint back into it's full value to enact on SRanipal Squeeze. To achieve this I will be putting Squint through an exponential curve in-which the curve will be influenced by Openness and EyeWide. 

I will play around with this approach and see if it achieves the results that I am after.

regzo2: Hmm not sure how it gets tracked on IPhone ARKit but from testing MeowFace I notice that Eye Squint gets canceled out once my eyes blink, which seems to be of a lesser fidelity than the Quest Pro's Lid Tightener and a complete inverse of SRanipal Eye Squeeze. Seems to be a limitation of this tracking interface rather than the standard. Still believe we should consolidate all of these into a single Squint parameter in the new standard.

ALVRTrackingInterface/ALVRTrackingInterface.cs
```diff
@@ -230,78 +246,86 @@ private void Update()
...
             // Recover true eye closed values; as you look down the eye closes.
             // from FaceTrackingSystem.CS from Movement Aura Scene in https://github.com/oculus-samples/Unity-Movement
             float eyeClosedL = Math.Min(1, expressions[(int)FBExpression.Eyes_Closed_L] + expressions[(int)FBExpression.Eyes_Look_Down_L] * 0.5f);
             float eyeClosedR = Math.Min(1, expressions[(int)FBExpression.Eyes_Closed_R] + expressions[(int)FBExpression.Eyes_Look_Down_R] * 0.5f);
 
             // Add Lid tightener to eye lid close to help get value closed
-            eyeClosedL = Math.Min(1, eyeClosedL + expressions[(int)FBExpression.Lid_Tightener_L] * 0.5f);
-            eyeClosedR = Math.Min(1, eyeClosedR + expressions[(int)FBExpression.Lid_Tightener_R] * 0.5f);
+            // eyeClosedL = Math.Min(1, eyeClosedL + expressions[(int)FBExpression.Lid_Tightener_L] * 0.5f);
+            // eyeClosedR = Math.Min(1, eyeClosedR + expressions[(int)FBExpression.Lid_Tightener_R] * 0.5f);
```
Adjerry91, line 292: Uncomment the EyeClosedL and EyeClosedR lines. EyeClosed for the Quest pro is more a confidence in that user has blinked so when closing the eyes slowly it is really hard to wink. The Lid Tightener, Squint, is required to help close the eyes. So this lines are important for proper eyelid tracking.

ALVRTrackingInterface/ALVRTrackingInterface.cs
```diff
@@ -230,78 +246,86 @@ private void Update()
...
+            public static float ChinRaiserTop = 1.0f;
+            public static float ChinRaiserBottom = 1.0f;
+            public static float MouthLowerDown = 1.0f;
+            public static float MouthUpperUp = 1.0f;
+            public static float MouthDimpler = 1.0f;
+            public static float MouthStretch = 1.0f;
+            public static float MouthPress = 1.0f;        //VERY NOT SENSITIVE
+            public static float MouthTightener = 1.0f;
+            public static float NoseSneer = 1.0f;
         }
```
Adjerry91, line 278: Is the scaling done in the UnifiedExpressions? If so may would need to have the scalers removed from the external module.

ALVRTrackingInterface/ALVRTrackingInterface.cs
```diff
@@ -342,149 +366,152 @@ private void UpdateEye()
...
+            unifiedExpressions[(int)UnifiedExpressions.BrowInnerUpRight].Weight = expressions[(int)FBExpression.Inner_Brow_Raiser_R] * TrackingSensitivity.BrowInnerUp;
+            unifiedExpressions[(int)UnifiedExpressions.BrowOuterUpLeft].Weight = expressions[(int)FBExpression.Outer_Brow_Raiser_L] * TrackingSensitivity.BrowOuterUp;
+            unifiedExpressions[(int)UnifiedExpressions.BrowOuterUpRight].Weight = expressions[(int)FBExpression.Outer_Brow_Raiser_R] * TrackingSensitivity.BrowOuterUp;
+
+            unifiedExpressions[(int)UnifiedExpressions.BrowOuterDownLeft].Weight = expressions[(int)FBExpression.Brow_Lowerer_L] * TrackingSensitivity.BrowDown;
+            unifiedExpressions[(int)UnifiedExpressions.BrowInnerDownLeft].Weight = expressions[(int)FBExpression.Brow_Lowerer_L] * TrackingSensitivity.BrowDown;
+            unifiedExpressions[(int)UnifiedExpressions.BrowOuterDownRight].Weight = expressions[(int)FBExpression.Brow_Lowerer_R] * TrackingSensitivity.BrowDown;
+            unifiedExpressions[(int)UnifiedExpressions.BrowInnerDownRight].Weight = expressions[(int)FBExpression.Brow_Lowerer_R] * TrackingSensitivity.BrowDown;
```
Adjerry91, line 422: Where should the combined parameters be for external modules? There was some combined brow parameters that were in this module. Will this be internal for Unified Expressions?

VRCFaceTracking/Params/Expressions/UnifiedExpressions.cs
regzo2: This equation does what I want:
EyeLidExpandedSqueeze (v1): (Openness * 0.8 + Widen * 0.2) - (1 - ((Openness^0.25)) * Squint)

Basically Squint won't affect the eye shape too much until the eyes start closing. Once the eyes are fully closed Squint will have full range to enact Squeeze. Should also fix the Quest Pro squinting problem for this specific parameter.

SRanipalExtTrackingModule/SRanipalTrackingInterface.cs
```diff
@@ -0,0 +1,363 @@
...
+                data.Shapes[(int)UnifiedExpressions.MouthUpperRight].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthUpperRight];
+                data.Shapes[(int)UnifiedExpressions.MouthLowerLeft].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthLowerLeft];
+                data.Shapes[(int)UnifiedExpressions.MouthLowerRight].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthLowerRight];
+
+                data.Shapes[(int)UnifiedExpressions.MouthSmileLeft].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthSmileLeft];
+                data.Shapes[(int)UnifiedExpressions.MouthSmileRight].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthSmileRight];
+                data.Shapes[(int)UnifiedExpressions.MouthFrownLeft].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthSadLeft];
+                data.Shapes[(int)UnifiedExpressions.MouthFrownRight].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthSadRight];
+
+                data.Shapes[(int)UnifiedExpressions.MouthRaiserUpper].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthLowerOverlay] - external.blend_shape_weight[(int)LipShape_v2.MouthUpperInside];
```
Adjerry91, line 321: Need to look at MouthRaiserUpper emulation as this tracking parameter I have need really tested on the Quest Pro

SRanipalExtTrackingModule/SRanipalTrackingInterface.cs
```diff
@@ -0,0 +1,363 @@
...
+
+                data.Shapes[(int)UnifiedExpressions.MouthUpperLeft].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthUpperLeft];
+                data.Shapes[(int)UnifiedExpressions.MouthUpperRight].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthUpperRight];
+                data.Shapes[(int)UnifiedExpressions.MouthLowerLeft].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthLowerLeft];
+                data.Shapes[(int)UnifiedExpressions.MouthLowerRight].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthLowerRight];
+
+                data.Shapes[(int)UnifiedExpressions.MouthSmileLeft].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthSmileLeft];
+                data.Shapes[(int)UnifiedExpressions.MouthSmileRight].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthSmileRight];
+                data.Shapes[(int)UnifiedExpressions.MouthFrownLeft].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthSadLeft];
+                data.Shapes[(int)UnifiedExpressions.MouthFrownRight].Weight = external.blend_shape_weight[(int)LipShape_v2.MouthSadRight];
```
Adjerry91, line 319: I do not see MouthStretch SRanipal to UnifiedExpression mapping. It usually triggered by strong frown expression.

ALVRTrackingInterface/ALVRTrackingInterface.cs
regzo2: There is a new Mutator class that will scale the values per-user with both user-definable values and a calibration sequence. These should be okay to leave but they may be redundant with the Mutator.

ALVRTrackingInterface/ALVRTrackingInterface.cs
regzo2: I believe we should only allow modules access to the base Unified Expression shapes, which modules have free-reign to send any transformed data they want to. All transforms of the base expressions into their combined counterparts are handled internally. My approach with the combined parameters is to try to retain the fidelity of the data and not to transmute the data like we were doing with V1 (SRanipal) combined parameters.

This architecture also lets us have a lot more flexibility in segregating modules from the internals of VRCFT.

SRanipalExtTrackingModule/SRanipalTrackingInterface.cs
regzo2: Oh looks like I accidentally included MouthUpperInside in this conversion. I think I will have MouthLowerOverlay slightly weighted to MouthUpperRaiser and fully weighted to MouthLowerRaiser to retain consistency

VRCFaceTracking/Params/Expressions/UnifiedExpressions.cs
Adjerry91: With the Quest Pro demo avatar be careful to avoid the corrective shapes with making the basis for the code. Here is my experience how the Quest Pro tracks squint and eye close:
- Eye close for Quest Pro is Non-linear, it is confidence of Blinking not closing so when blinking it is strong value of 1
- Closing the eyes slow is very insensitive when closing will only get ~0.5 raw value
- Squint triggers strongly when closing the eyes slowly and is mostly saturated when you completely close the eyes. There is no more tracking squint once the eye is closed and you squeeze your eye.
- I have been using squint to help close the eye for the Quest Pro module by add the value to eye closed. You can see previous comment to bring the code back.

regzo2:
Oh also for the calibrator I think I will add a deviation variable to indicate how far you can deviate from the raw value for each calibration value. There are definitely some parameter that I want to make sure have pretty good consistency (like JawOpen, Mouth), and some that might have be more subjective. Basically modules can set a deviation that they expect and VRCFT will try to calibrate within that deviation, and maybe VRCFT could calculate one if needed. Of this would rely on modules actually sending good data too so should be pretty trusting on the data being sent, and this would only be applied if the calibrator is enabled (which is disabled by default as I would want it to be an opt-in feature for modules and users).

regzo2:
@Adjerry91  Could you try this build and see if the Quest Pro emulates into SRanipal properly or as expected?

regzo2:
Looks like Mutator functions are partially broken.

VRCFaceTracking/Params/Expressions/UnifiedExpressions.cs
regzo2: > With the Quest Pro demo avatar be careful to avoid the corrective shapes with making the basis for the code

I think in this case and many other cases the overlapping shape corrections are an indication of their intention to potentially have the output be able to do those shapes in the future. Basically we should have our standard be capable of that level of fidelity but not really enforce tracking standards to track in that way.

> * Squint triggers strongly when closing the eyes slowly and is mostly saturated when you completely close the eyes. There is no more tracking squint once the eye is closed and you squeeze your eye.

This seems to be a similar behavior that AR52 (specifically in this case MeowFace) has as well, where specifically lifting your lower eyelid and blinking can affect it. Avatars should be built to accept any sort of mixing of these parameters, and in the case of the Quest Pro demos they seem to be able to accept that kind of mixing of data. Squint in Unified will treat AR52/Meta FACS Squint appropriately as well as SRan Eye Squeeze, as this can be considered a tracking limitation of the interfaces in question and should still interface into our current standard properly with intended tracking.

> * I have been using squint to help close the eye for the Quest Pro module by add the value to eye closed. You can see previous comment to bring the code back.

I don't have a Quest Pro to test with but I do remember this was something that was done in the Quest Pro demos when looking at the source scripts. I have reverted this behavior.

ALVRTrackingInterface/ALVRTrackingInterface.cs
regzo2: Okay I'm looking through the code again and after implementing the reformulated v1 (SRanipal) EyeLidExpandedSqueeze combined parameter this lid tightener addition shouldn't be needed as adding lid tightener into eye openness will most likely cause issues with use in the new standard. 

I am using MeowFace and it works exactly as expected when using the new reformulated v1 parameter (I was having the same issue you were having with Quest Pro where the eye closes earlier than expected), the new formula will allow Squint to affect the eye blink and once the eyes are closed it will cause squint to be much more useable (which retains the behavior of both SRanipal eye squeeze and FACS/AR52 squint).

If you build an avatar to work with the new standard using the new parameters you also shouldn't have to worry about having the eyelid closing effectively as Squint will be completely separated from blinking and eye widen.

I'm just wondering if this addition of Lid Tightener is there because of the old parameters or if it's actually needed for the new standard. From my own testing it shouldn't be necessary.

SRanipalExtTrackingModule/SRanipalTrackingInterface.cs
regzo2: Also this will only affect the new standard parameters and won't affect the old parameters at all.

ALVRTrackingInterface/ALVRTrackingInterface.cs
regzo2: >// Lid Tightener is not tracked the same as SRanipal eye squeeze. This causes problems with combined parameters. The lid tightener has more controls the fine state of closing the eye while the eye lid is more of control blinking.
            // Eye close is non-linear and seems to be based on the confidence of that eye blink is detected. Lid tightener will be used to control the eye state thus squeeze will be disabled for now for the Quest Pro mapping.
            
These comments seem to indicate that it was for originally mapping into SRanipal base parameters. The combined parameters are the reason that SRanipal eye squeeze doesn't track as FACS/AR52 squint does. Mapping to the new standard should allow tightener to function as expected, especially with the reformulated EyeLidExpandedSqueeze.

VRCFaceTracking/Params/Expressions/UnifiedExpressionsMerger.cs
Adjerry91: Still needs fixed

ALVRTrackingInterface/ALVRTrackingInterface.cs
Adjerry91: Need to add add the combined parameters
- BrowDownLeft
- BrowDownRight
- BrowsDown
- BrowsInnerUp
- BrowsOuterUp

VRCFaceTracking/Params/Expressions/Legacy/Eye/EyeTrackingParams.cs
```diff
@@ -58,18 +56,18 @@ public static class EyeTrackingParams
...
                 "RightEyeLidExpandedSqueeze", 0.5f, true),
             new EParam(exp =>
-                (((exp.Shapes[(int)UnifiedExpressions.EyeWideLeft].Weight * .2f + exp.Eye.Left.Openness * .8f) + 
-                (exp.Shapes[(int)UnifiedExpressions.EyeWideLeft].Weight * .2f + exp.Eye.Left.Openness * .8f)) / 2.0f) -
-                (((float)((1.0f - Math.Pow(exp.Eye.Left.Openness, .36)) * exp.Shapes[(int)UnifiedExpressions.EyeSquintLeft].Weight) + 
-                (float)((1.0f - Math.Pow(exp.Eye.Right.Openness, .36)) * exp.Shapes[(int)UnifiedExpressions.EyeSquintRight].Weight)) / 2.0f), 
+                (exp.Shapes[(int)UnifiedExpressions.EyeWideLeft].Weight * .2f +
+                0.8f *exp.Eye.Left.Openness - exp.Shapes[(int)UnifiedExpressions.EyeSquintLeft].Weight) + 
+                (exp.Shapes[(int)UnifiedExpressions.EyeWideRight].Weight * .2f +
+                0.8f * exp.Eye.Right.Openness - exp.Shapes[(int)UnifiedExpressions.EyeSquintRight].Weight) / 2.0f,
```
regzo2, line 70: Not entirely sure why these are being reverted, you are going to run into the issue of Squint causing the eyelids to close pre-maturely. I am trying to have it so that squint doesn't affect the eyelids until the eyes are very nearly closed, this reversion will cause issues with 'proper' Unified implementations. It's just simply a glaring oversight with these ExpandedSqueeze parameters having Squint contending with 2 other parameters. It was built with the limitations of SRanipal in mind, and our v2 parameters will make sure that this combination isn't included. Also keep in mind the compensation won't apply to other eyelid params, even in the legacy params (EyeLidExpanded and the other uncombined ones).

Adjerry91: I am confused where the compensation should be handled at, I did it in ALVR tracking module for reducing the squint with the openness. Should this be handled in the module as it is sensitivity issue with multiple parameters? just removed all the power element with combined parameters.

ALVRTrackingInterface/ALVRTrackingInterface.cs
```diff
@@ -270,25 +272,22 @@ private void UpdateEyeData(ref UnifiedEyeData eye, ref float[] expressions)
...
             #endregion
-        }
 
-        private void UpdateEyeExpressions(ref UnifiedExpressionShape[] unifiedExpressions, ref float[] expressions)
-        {
             #region Eye Expressions Set
 
-            unifiedExpressions[(int)UnifiedExpressions.EyeWideLeft].Weight = expressions[(int)FBExpression.Upper_Lid_Raiser_L] ;
-            unifiedExpressions[(int)UnifiedExpressions.EyeWideRight].Weight = expressions[(int)FBExpression.Upper_Lid_Raiser_R] ;
+            unifiedExpressions[(int)UnifiedExpressions.EyeWideLeft].Weight = (float)Math.Pow(1.0f - expressions[(int)FBExpression.Eyes_Closed_L], 1.0f / EyeLidWeight) * (float)Math.Pow(expressions[(int)FBExpression.Upper_Lid_Raiser_L], 1.0 / WidenWeight);
```
regzo2, line 286: Same as mention as above, Widen does not need any compensation.

regzo2: Basically for this module the *only* parameter that needs to be compensated for is Openness, and AFAIK only lid tightener and eye close should be the only ones being weighted into Openness. Every other parameter should be 1:1. We shouldn't expect every module to compensate for v1 parameters at all.

regzo2: If you're okay with that I am going to make some changes to revert some of the code back, and feed Openness like what I am doing with the MeowFace module. I will formulate the ExpandedSqueeze to work best with that approach.

VRCFaceTracking/Params/Expressions/Legacy/Eye/EyeTrackingParams.cs
regzo2: Only openness needs compensation, none of the other parameters will need it from ALVR module. This specific parameter type (ExpandedSqueeze) is the only one going to have the formula it has simply because it has squeeze and widen included in it, which are actively fighting against how the eye is being opened for this specific parameter. The ramp is there to prevent squeeze from affecting the shape until the eyes are closed (which will work around both the SRanipal tracking and AR52/FACS tracking). I think the ramp could be altered to be a lot more strong now that I have a clear idea on how the data should be sent from modules.

The only way I can explain how you should send the data from the modules is treat it like your targeting the values to be sent to the Unified (v2) parameters only (in which case you only need to use Lid Tightener and Eye Closed to control the Openness parameter like I did in the MeowFace formula), and after it works well for v2 then the SRanipal emulator will be developed to correct those parameters to work well for the old v1 (SRanipal) standard.

Hope that makes it clear, basically build the module to target v2 parameters and the SRanipal emulator will convert it to work for the old v1 parameters.

ALVRTrackingInterface/ALVRTrackingInterface.cs
Adjerry91: That will work

VRCFaceTracking/OSC/OSCMain.cs
```diff
@@ -70,8 +70,36 @@ private void Recv()
...
+                    }
+                    break;
+                case "/avatar/parameters/LipTrackingActive":
+                    if (UnifiedLibManager.ExpressionStatus != ModuleState.Uninitialized)
+                    {
+                        if (!(bool)newMsg.Value)
+                            UnifiedLibManager.ExpressionStatus = ModuleState.Idle;
+                        else UnifiedLibManager.ExpressionStatus = ModuleState.Active;
+                    }
+                    break;
```
PinballsWizard, line 93: If `LipTrackingActive` and `ExpressionTrackingActive` are to be handled identically, would it make sense to remove this and let the case fall through into the next?

regzo2: They are both meant to change the state of the Lip/Expression module so they could both fall into the same case.

regzo2: Changed in https://github.com/benaclejames/VRCFaceTracking/pull/109/commits/e8bc944abd2a9c15d890c1ad6cae5e14462167c5

regzo2:
As far as moving to merging this into master, what do you all think we should do with as the modules included with the repo atm? VRCFT after this merge won't have any SRanipal dependencies so it won't be necessary to have the SRanipal module on this repo. Should we move the modules to their own repo(s)?

SRanipalExtTrackingModule/SRanipalExtTrackingModule.csproj.user
```diff
@@ -0,0 +1,14 @@
...
+  <PropertyGroup>
+    <ProjectView>ProjectFiles</ProjectView>
+  </PropertyGroup>
+  <PropertyGroup>
+    <ReferencePath>
+    </ReferencePath>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|AnyCPU'">
+    <StartAction>Project</StartAction>
+    <StartProgram>C:\Users\lazlo\source\repos\VRCFaceTracking\VRCFaceTracking\bin\Release\VRCFaceTracking.exe</StartProgram>
```
benaclejames, line 12: We might wanna remove this `StartProgram` block, or the file itself as it contains somewhat sensitive data and is autogenerated

VRCFaceTracking/Assets/UI/App.xaml.cs
```diff
@@ -10,8 +12,37 @@ public partial class App
...
-            BindingOperations.EnableCollectionSynchronization(Logger.ConsoleOutput, Logger.ConsoleLock);
-            _oscThread.Start();
+            try
+            {
+                BindingOperations.EnableCollectionSynchronization(Logger.ConsoleOutput, Logger.ConsoleLock);
+                _oscThread.Start();
+            }
+            catch ( Exception e ) 
+            {
+                string filePath = Utils.PersistentDataDirectory + "/ExceptionLog.txt";
```
benaclejames, line 22: We should try and move this to the logger in the future, but fine for now as the file logging functionality is yet to be fully implemented. Just something to keep in mind 👍

VRCFaceTracking/Library/UnifiedLibManager.cs
```diff
@@ -0,0 +1,296 @@
...
+                catch (FileLoadException)
+                {
+                    Logger.Warning(dll + " Assembly could not be loaded. Skipping.");
+                    continue;
+                }
+                catch (Exception e)
+                {
+                    Logger.Warning(e.Message + " Assembly not able to be loaded. Skipping. ");
+                    continue;
+                }
```
benaclejames, line 192: Feels like this would be the only exception we needed if we printed the type of exception as well as the message. The names are pretty self-explanatory though feel free to leave this as-is if you disagree

VRCFaceTracking/Logger.cs
```diff
@@ -11,31 +11,60 @@ public static class Logger
...
+            try
+            {
+                ConsoleOutput.Add(new Tuple<string, string>(formattedStr, "White"));
+                Console.WriteLine(formattedStr);
+            }
+            catch (Exception e)
+            {
+                ConsoleOutput.Add(new Tuple<string, string>(formattedStr + "Message intercepted while writing. " + e, "White"));
+                Console.WriteLine(formattedStr);
+            }
```
benaclejames, line 26: Just out of curiosity, how often does this happen and what are the conditions for the catch to be triggered?

VRCFaceTracking/OSC/OSCMain.cs
```diff
@@ -70,8 +51,29 @@ private void Recv()
...
+                case "/avatar/parameters/LipTrackingActive":
+                case "/avatar/parameters/ExpressionTrackingActive":
+                    if (UnifiedLibManager.ExpressionStatus != ModuleState.Uninitialized)
+                    {
+                        if (!(bool)newMsg.Value)
+                            UnifiedLibManager.ExpressionStatus = ModuleState.Idle;
+                        else UnifiedLibManager.ExpressionStatus = ModuleState.Active;
+                    }
+                    break;
+            }
```
benaclejames, line 76: I wonder if we could replace this in exchange for another class for handling OSC toggle types derived from a simple `BoolBaseParam`?

VRCFaceTracking/OSC/OSCMessage.cs
```diff
@@ -109,17 +113,61 @@ public OscMessage(byte[] bytes)
...
+                    break;
+                case 91: // OSC Type tag: '[' ; Represents start of an array.
+                    Logger.Error("Unimplemented Array Type detected: " + type + " for name " + Address);
+                    break;
+                case 93: // OSC Type tag: ']' ; Represents end of an array.
+                    Logger.Error("Unimplemented Array Type detected: " + type + " for name " + Address);
+                    break;
+
+                #endregion
+
```
benaclejames, line 167: https://user-images.githubusercontent.com/11042095/222858150-0530b601-1b66-40e1-885e-f201259b0f89.mp4


Just GOTO default I beg you 🤣

benaclejames (changes requested):
Looking good! Just a few minor things as listed in the review, though I've intentionally left the ALVR modules unreviewed as I have no idea what I'm talking about there (CC @Adjerry91). 
Really excited to get things merged, and thanks again to everyone who contributed!

VRCFaceTracking/Logger.cs
regzo2: I've seen it happen when switching out external modules and reinitializing while the logger is handling a message from an external module. I believe the changes you've made in a different commit to the logger are a more thorough fix to that issue.

VRCFaceTracking/SDK/Attributes.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <summary>
+        /// Container of the type of data the module is expected to send.
+        /// </summary>
+        public VRCFTModuleType ModuleType { get; private set; }
+
+        public VRCFTModuleTypeAttribute(VRCFTModuleType moduleType)
+        {
+            ModuleType = moduleType;
+        }
+    }
```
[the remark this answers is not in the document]
regzo2: Currently this specific attribute isn't used by any system within VRCFT as it is already handled by the ExtTrackingModule class parsing an assembly. You are right though this specific attribute is redundant as supported module types (Eye, Expression (Lip)) modules are already handled within ExtTrackingModule. The load order oriented attributes are handled by the assembly loader in the cases of ordering the load order of modules for some edge cases.

VRCFaceTracking/Assets/UI/MainWindow.xaml.cs
```diff
@@ -169,10 +187,75 @@ private void MainWindow_OnSizeChanged(object sender, EventArgs eventArgs)
...
+
+                Logger.Msg("Calibrating deep normalization for 30s.");
+                Thread.Sleep(30000);
+
+                UnifiedTracking.Mutator.CalibrationWeight = 0.2f;
+                Logger.Msg("Fine-tuning normalization. Values will be saved on exit.");
+                
+            });
+            _thread.Start();
+        }
```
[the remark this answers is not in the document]
regzo2: Can do! I had this in the UI space as I was expecting that the UI would handle a lot of the aspects of controlling the underlying systems of VRCFT, but I agree that having these within the Mutator would be more obvious and accessible in the codebase.

VRCFaceTracking/Assets/UI/App.xaml.cs
regzo2: Sounds good, I believe I did this really early on in the PR and neglected to finalize it further.

VRCFaceTracking/OSC/OSCMessage.cs
regzo2: Yeeeeeaaaahhhh should default the unimplemented ones. Got a little too ambitious with the changes my b 😥.

VRCFaceTracking/Library/UnifiedLibManager.cs
regzo2: I agree, these were meant to be more concise messages but definitely having the exception message is probably the most useful for troubleshooting code.

SRanipalExtTrackingModule/SRanipalExtTrackingModule.csproj.user
regzo2: Should remove it since it doesn't affect building the project.

VRCFaceTracking/OSC/OSCMain.cs
regzo2: Maybe we could flesh out this BoolBaseParam derivitive I added to work around the LipTrackingActive and EyeTrackingActive parameters, and allow it to accept unique func delegates that can then be called, or it could be it's own class:  [ConditionalBoolParameter](https://github.com/benaclejames/VRCFaceTracking/pull/109/files/5d7d16c561782b3f5c037707cd71b3e03d157d01#diff-ac24c6fa2cf66c9ea74377c07d568956a9bf661364132cc419604b0b9d5cb07eR43)?

It could be set in it's own container (eg. Toggles) that are only used to change states within VRCFT and not intended for output parameters, since we aren't really expecting output parameters to be controlled via recieves.

VRCFaceTracking/Assets/UI/MainWindow.xaml.cs
```diff
@@ -169,10 +187,75 @@ private void MainWindow_OnSizeChanged(object sender, EventArgs eventArgs)
             } 
         }
 
+        private void MainWindow_Loaded(object sender, EventArgs eventArgs)
+        {
+            // List box containing all selectable modules that implement ExtTrackingModule.
+            UnifiedLibManager.ReloadModules();
```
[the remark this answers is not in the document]
regzo2: I can move this back to MainStandalone. I had this here to work around an issue with the UI not having the listbox with all available modules, I agree having this Reload call in the UI is not ideal.

VRCFaceTracking/OSC/OSCMessage.cs
regzo2: Resolved in https://github.com/benaclejames/VRCFaceTracking/pull/109/commits/da0730058c64d099efb15ac120645fa0e60ff468

VRCFaceTracking/Library/UnifiedLibManager.cs
regzo2: Removed the extra catches and kept the messaged exception in https://github.com/benaclejames/VRCFaceTracking/pull/109/commits/6c285efaa0bc322e1f46db9842062a7558e432ae

VRCFaceTracking/Assets/UI/MainWindow.xaml.cs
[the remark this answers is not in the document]
regzo2: Moved Reload to MainStandalone in [a8fa35f](https://github.com/benaclejames/VRCFaceTracking/pull/109/commits/a8fa35f895424e259428f3316ac0003a882eda1e)

VRCFaceTracking/Assets/UI/MainWindow.xaml.cs
```diff
@@ -169,10 +187,75 @@ private void MainWindow_OnSizeChanged(object sender, EventArgs eventArgs)
...
+
+        private void EnableSmoothing_Checked(object sender, RoutedEventArgs e)
+        {                   
+            UnifiedTracking.Mutator.SmoothingMode = true;
+        }
+
+        private void EnableSmoothing_Unchecked(object sender, RoutedEventArgs e)
+        {
+            UnifiedTracking.Mutator.SmoothingMode = false;
+        }
```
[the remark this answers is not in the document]
regzo2: Resolved in [86ccef0](https://github.com/benaclejames/VRCFaceTracking/pull/109/commits/86ccef0f1ee4dc2d8e9e73063ccc2bfae39b853b)

VRCFaceTracking/Params/Expressions/Legacy/Lip/LipShapeMerger.cs
```diff
@@ -0,0 +1,151 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using VRCFaceTracking.Params.Lip;
+
+namespace VRCFaceTracking.Params.LipMerging
+{
+    [Obsolete("SRanipal Lip Shapes are being phased out. Please switch to Unified Expressions.")]
```
regzo2, line 8: I noticed that this was tagged as Obsolete, generally module devs don't touch this class when sending tracking data to external modules anyhow. @benaclejames what do you think about labelling the other SRanipal classes (such as the SRanipal enum, and shape transforms of SRanipal) and shapes as Obsolete?

VRCFaceTracking/Params/Expressions/UnifiedExpressions.cs
```diff
@@ -0,0 +1,235 @@
...
+        Max
+    }
+
+    /// <summary> 
+    /// Represents the type of Legacy Shape data being sent by UnifiedExpressionData, in the form of enumerated SRanipal shapes.
+    /// </summary>
+    /// <remarks>
+    /// This enum is not intended to be used directly by modules in the final iteration, and instead will be used as a compatibility layer for making the new Unified system backwards compatible with older VRCFT avatars.
+    /// </remarks>
+    internal enum SRanipal_LipShape_v2
```
regzo2, line 193: This is currently labelled as internal (a change I did near the start of this PR), but we could re-expose it and make sure that it's named the same as before in VRCFT with the obsolete attr.

VRCFaceTracking/Params/Expressions/Legacy/Lip/LipShapeConversion.cs
```diff
@@ -0,0 +1,90 @@
+﻿using System.Linq;
+
+namespace VRCFaceTracking.Params.Lip
+{
+    internal interface ICombinedShape
```
regzo2, line 5: All shapes that implement ICombinedShape would become obsolete too since only the old SRanipal shapes use it, (Unified Expressions' combined shapes is purely controlled with delegates and doesn't touch this class)

regzo2 (commented):
Noticed some other classes that could have the Obsolete tag

VRCFaceTracking/Assets/UI/MainWindow.xaml.cs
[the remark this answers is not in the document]
regzo2: Resolved in https://github.com/benaclejames/VRCFaceTracking/pull/109/commits/0bea7eb9b267918b26259a82d18405410ab8ceaa

SRanipalExtTrackingModule/SRanipalExtTrackingModule.csproj.user
regzo2: Removed in https://github.com/benaclejames/VRCFaceTracking/pull/109/commits/1c56093db19ef05654b4880307512d69920fc441

VRCFaceTracking/Logger.cs
regzo2: Gonna mention https://github.com/benaclejames/VRCFaceTracking/issues/121 for future ref.

VRCFaceTracking/Params/Expressions/Legacy/Lip/LipShapeMerger.cs
benaclejames: Hmm, provided they're within the VRCFT namespace, that sounds like a good idea!

VRCFaceTracking/Params/Expressions/UnifiedExpressions.cs
benaclejames: IMO we shouldn't be redefining classes at all. If its not something we use in the VRCFT namespace, remove it. If we do use it in VRCFT, we should remove all other instances of it from included external tracking modules

regzo2: Currently this enum is only used by the SRanipal conversion and the obsoleted LipMerger combined parameters, otherwise it's not used outside of that and this one shouldn't be used by third-party modules, keeping it internal is probably best. Was just wondering how you wanted to go about exposing VRCFT stuff 😎

VRCFaceTracking/Params/Data/UnifiedTrackingMutator.cs
```diff
@@ -0,0 +1,146 @@
...
+namespace VRCFaceTracking
+{
+    /// <summary>
+    /// Container of all functions and structures retaining to mutating the incoming Expression Data to be usable for output parameters.
+    /// </summary>
+    public class UnifiedTrackingMutator
+    {
+        public struct Mutation
+        {
+            public string Name;
```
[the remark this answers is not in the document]
regzo2: Once the UI can allow mutator adjustments per parameter this won't be necessary at all as an addendum, the current Mutator system just needs more exposure to the UI.

VRCFaceTracking/SDK/Attributes.cs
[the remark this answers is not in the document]
regzo2: I can comment this out for now, we can discuss if we want to fully adopt using reflected attributes for more stuff in external modules in the future.

VRCFaceTracking/Params/Expressions/Legacy/Lip/LipShapeMerger.cs
regzo2: Ah that makes sense, given that most of the SRanipal stuff was either removed as dependencies or replaced with a VRCFT namespaced version (specifically SRanipal_LipShape_v2 in UnifiedExpressions.cs), this specific class being obsoleted makes sense. I think that clears up everything!

VRCFaceTracking/SDK/Attributes.cs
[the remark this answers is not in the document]
regzo2: Resolved in https://github.com/benaclejames/VRCFaceTracking/pull/109/commits/e6d416684df4e92e1682606fcdfa47a134fe7210 and https://github.com/benaclejames/VRCFaceTracking/pull/109/commits/946b8c572c1d30f05af8032b32de871f069808e2

benaclejames:
I think I spoke about this with @regzo2 at some point in the past, but how much of the interface is required for new users? My main questions currently are
1) Is the calibration necessary on boot?
2) Is module selection required for initial setup?
3) In what situations is the smooth expressions slider necessary?

regzo2:
None of the added features are necessary for using VRCFT, it will work exactly as before they were introduced.

> 1. Is the calibration necessary on boot?

Calibration is not necessary at all to use VRCFT, when it's disabled the data being sent into VRCFT is directly from modules. Calibration is toggled and any calibrated data is saved (the Calibrator toggle completely disables any calibration from actually being applied, and is off by default), and is meant to be a per-user adjusted optionally (currently the UI has it a bit under-exposed for ease of use but the underlying system has the ability to calibrate and adjust each parameter).

> 2. Is module selection required for initial setup?

Modules are automatically loaded and initialized in the order they are first loaded (alphabetical); Users can pick a load order which is then saved for later.

> 3. In what situations is the smooth expressions slider necessary?

Smoothing is optional as well, there are some interfaces that might benefit from it but it shouldn't necessary for that use case in general. A good example of where it might be really nice for that use case is for the Focus 3 tracking from the SRanipal module, which has no smoothing applied for some reason. It's really just there if users want their expressions to be more smooth (it's similar to how other apps like VSeeFace let you set smoothing on expressions). It's not necessary from a technical standpoint, more-so from a subjective user or aesthetic use-case.

benaclejames:
Sweet, thanks for clearing that up. I'll be sure to include extracts from that in the release notes when this is eventually out, it's gonna be a big changelog haha

Given that this is merging into the quest pro branch and the low numbers of vive focus 3 users, would it be possible to remove/hide the smoothing slider for this update until we find a better way to genericise what options modules are allowed/able to expose to the end user with a future UI update perhaps?

That being said, thank you so much again for everything you've done here! I think after the slider is hidden, I'll be happy to merge!

regzo2:
Yeah I can definitely hide it from the front-end until we discuss what we want to do in the finalized interface, but just to make it clear it's purely meant to be a globally set user adjustable thing per parameter, not necessarily tied to a module implementation (although I believe that would be a really interesting approach for it). I'll commit hiding it from the UI and we can figure out how exactly we want to go about doing parameter smoothing in VRCFT.

benaclejames (approved):
Looks good! Merge time!

PR #110, opened 2022-11-23
External module loading order.

ULemon:
Try to use external module first, instead of the stock module. If a PC has already installed SRanipal, it will always load the stock module. Just like how og DOOM loads WADs.


benaclejames:
This is intentional behaviour. If a module is built into the app, chances are it's important. If you don't have SRanipal installed, or do not have a compatible headset, it will fail and fall back to the next external module anyways. This will be fixed in the next update when SRanipal is moved to an external module as well

PR #112, opened 2022-11-27
styling consistency

Fox-McCloud:
Minor styling consistency for `MouthUpperInsideOverturn` and `MouthLowerInsideOverturn`.

Doubt this is the cause of the two to be non-functional, but none the less, this brings them in-line with the the code here.


benaclejames:
Hmm, these seemed to have been fixed at some point, sorry for not merging this sooner but thank you for making us aware of the issue 😅

Issue #115, opened 2022-12-21
Pimax eye tracking support is gone?


Kieaer:
Looking at PimaxEyeTracking in the README, this is a MelonLoader only mod.

Kieaer:
```
[12:07:42] VRCFT Standalone Initializing!
[12:07:42] Loaded library C:\Users\cloud\AppData\Roaming\VRCFaceTracking\StockLibs\libHTC_License.dll
[12:07:42] Loaded library C:\Users\cloud\AppData\Roaming\VRCFaceTracking\StockLibs\nanomsg.dll
[12:07:42] Loaded library C:\Users\cloud\AppData\Roaming\VRCFaceTracking\StockLibs\SRWorks_Log.dll
[12:07:42] Loaded library C:\Users\cloud\AppData\Roaming\VRCFaceTracking\StockLibs\ViveSR_Client.dll
[12:07:42] Loaded library C:\Users\cloud\AppData\Roaming\VRCFaceTracking\StockLibs\SRanipal.dll
[12:07:42] Initialized DependencyManager Successfully
[12:07:42] Initialized UnifiedLibManager Successfully
[12:07:42] Finding and initializing runtimes...
[12:07:42] Loading External Modules...

Unhandled Exception: System.ArgumentException: Directory in URL is invalid.
   위치: System.Security.Util.DirectoryString.CreateSeparatedString(String directory)
   위치: System.Security.Util.URLString.ParseFileURL(String url)
   위치: System.Security.Util.URLString.GetFileName()
   위치: System.Reflection.RuntimeAssembly.InternalLoadAssemblyName(AssemblyName assemblyRef, Evidence assemblySecurity, RuntimeAssembly reqAssembly, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
   위치: System.Reflection.RuntimeAssembly.InternalLoadFrom(String assemblyFile, Evidence securityEvidence, Byte[] hashValue, AssemblyHashAlgorithm hashAlgorithm, Boolean forIntrospection, Boolean suppressSecurityChecks, StackCrawlMark& stackMark)
   위치: System.Reflection.Assembly.LoadFrom(String assemblyFile)
   위치: VRCFaceTracking.UnifiedLibManager.LoadExternalModules()
[...]

PR #116, opened 2022-12-21
Fix can't load external library

Kieaer:
image
It will solve #115 
may be.


benaclejames (commented):
Both of these methods execute the same underlying code for file discovery. Namely `FileSystemEnumerableIterator::ctor`. I'm skeptical this would actually change any load functionality at all 🤔

Kieaer:
@benaclejames Directory browsing is slightly slower, but it works unlike GetFiles.
I don't know why.

benaclejames:
Our module loading system has significantly changed and this no longer seems to be an issue.
Closing as stale PR

PR #118, opened 2023-01-09
Fix invalid osc config causing crashes

kusomaigo:
A few users had experienced crashes when starting VRChat or switching in avatars. It turns out if there is a malformed or empty .json in the avatar files it will cause an Exception during config parsing.


benaclejames:
No longer required as we have a settings tab to configure OSC.
Closing as stale

PR #122, opened 2023-03-10
Merge quest-pro into main ready for release

### LINKED ISSUE #106: Request: Support for Meta Quest Pro

benaclejames:
Pretty self explanatory, merge conflicts need fixing but a final go-ahead from the main contributors would be wonderful in the meantime

Closes #106


regzo2:
There are a few things that we need to make sure are retained as the conflicts get merged that are already a part of main, and might not necessarily show up as conflicts b/c of the file restructure (as of Nov 5, 2022):
1. https://github.com/benaclejames/VRCFaceTracking/commit/fc76bf9dc2b4492664b013cd82de860340f8d68b - disable-args
2. https://github.com/benaclejames/VRCFaceTracking/commit/5a813a0f6506c654bce9556c4dbe1dc3b21afecb - removed Zone identifiers (fixes the unblocking of modules)
3. https://github.com/benaclejames/VRCFaceTracking/commit/0865da9e20c3ff1748e4368c78feaf15c75beb5a - possibly these changes to UnifiedLibManager (from disable-args changes)?
4. Any other commits I can't think of lol.

regzo2 (commented):
Here are a huge amount of comments that kind of gloss over some key points with the changes I made, some pain points and some changes I have noticed as well when working on the overhaul. Most all of the comments are not necessarily intended for change immediately, more-so just for focus of discussion. 

Any changes not mentioned here are either file-structure changes or are pretty obvious changes (like the UnifiedTracking changes for eg.), the listed comments are more subtle or niche changes and points.

If anyone needs a starting point then starting with UnifiedTracking is where the main module interface starts, followed with UnifiedLibManager and the Expressions folder which contains all the base and merged parameters (UnifiedExpressions, Merger, and Legacy parameters).

Anyone feel free to comment on the reviews I made!

SRanipalExtTrackingModule/SRanipalTrackingInterface.cs
```diff
@@ -0,0 +1,416 @@
...
+        private static bool eyeEnabled = false, lipEnabled = false;
+
+        private static CancellationTokenSource _cancellationToken;
+        
+        public override (bool SupportsEye, bool SupportsExpressions) Supported => (true, true);
+
+        public override (bool eyeSuccess, bool expressionSuccess) Initialize(bool eyeAvailable, bool expressionAvailable)
+        {
+            // Look for SRanipal assemblies here. Placeholder for unmanaged assemblies not being embedded in the dll.
+            Directory.SetCurrentDirectory(Utils.CustomLibsDirectory + "\\ModuleLibs");
```
regzo2, line 30: This looks in a subfolder for the SRanipal libs, though if we plan on embedding it into the SRanipal module maybe this won't be needed?

VRCFaceTracking/Params/Expressions/Legacy/Lip/UnifiedSRanMapper.cs
```diff
@@ -0,0 +1,101 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace VRCFaceTracking.Params.Lip
+{
+    internal static class UnifiedSRanMapper
```
regzo2, line 6: Just want to make sure that the SRanipal conversion works exactly as intended or works as expected for everyone? SRanipal should work exactly as it did before the parameter overhaul (VRCFT 4.0), and it should be better suited to handle Unified Expressions.

VRCFaceTracking/Params/Expressions/UnifiedExpressions.cs
```diff
@@ -0,0 +1,235 @@
+﻿namespace VRCFaceTracking.Params
+{
+    // TODO: Create expressions map to relate legacy shapes to new ones
+
+    /// <summary> 
+    /// Represents the type of Shape data being sent by UnifiedExpressionData, in the form of enumerated shapes.
+    /// </summary>
+    /// <remarks>These shapes have a strong basis on the underlying muscular foundations of the entire head including the face, eyes, tongue, and inner mouth expressions.</remarks>
+    public enum UnifiedExpressions
```
regzo2, line 9: There are several UnifiedExpressions shapes that go completely unused by any interface, as follows:
```
NasalDilationRight, // Right side nose's canal dilates. Basis on the alar nasalis muscle.
NasalDilationLeft, // Left side nose's canal dilates. Basis on the alar nasalis muscle.
NasalConstrictRight, // Right side nose's canal constricts. Basis on the transverse nasalis muscle.
NasalConstrictLeft, // Left side nose's canal constricts. Basis on the transverse nasalis muscle.

JawBackward, // Pulls the jawbone backwards slightly. Based on the retraction of the temporalis muscle.
JawClench, // Specific jaw muscles that forces the jaw closed. Causes the masseter muscle (visible close to the back of the jawline) to visibly flex.
JawMandibleRaise, // Raises mandible (jawbone).

TongueBendDown, // Pushes tip of the tongue below the rest of the tongue in an arch.
TongueCurlUp, // Pushes tip of the tongue above the rest of the tongue in an arch.
TongueSquish, // Tongue becomes thinner width-wise and slightly thicker height-wise.
TongueFlat, // Tongue becomes thicker width-wise and slightly thinner height-wise.

TongueTwistRight, // Tongue tip rotates clockwise from POV with the rest of the tongue following gradually.
TongueTwistLeft, // Tongue tip rotates counter-clockwise from POV with the rest of the tongue following gradually.

SoftPalateClose, // Visibly lowers the back of the throat (soft palate) inside the mouth to close off the throat.
ThroatSwallow, // Visibly causes the Adam's apple to pull upward into the throat as if swallowing.

NeckFlexRight, // Flexes the Right side of the neck and face (causes the right corner of the face to stretch towards.)
NeckFlexLeft, // Flexes the left side of the neck and face (causes the left corner of the face to stretch towards.)

```

However as far as using them on avatars I left that as a possibility if avatar creators wanted to have the shapes on their avatars for potential future interfaces. What do you all think about having these documented and, if exposed in documentation, how relevant should they be? They are heavily based on facial anatomies and other standards implementing them (such as FACS and additions to it). 

Or really any change to them I'm pretty open about it 😅

VRCFaceTracking/Params/Expressions/UnifiedSimpleExpressions.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        BrowDownRight, // Lowers and pinches the right eyebrow.
+        BrowDownLeft, // Lowers and pinches the left eyebrow.
+        MouthSmileRight, // Moves the right corner of the lip into a smile expression.
+        MouthSmileLeft, // Moves the left corner of the lip into a smile expression.
+        MouthSadRight, // Moves the right corner of the lip into a sad expression.
+        MouthSadLeft, // Moves the left corner of the lip into a sad expression.
+        MouthUpperUpRight, // Raises the entire upper right of the lips.
+        MouthUpperUpLeft, // Raises the entire upper left of the lips.
+    }
```
regzo2, line 26: I added these to be more recognizable and simplified expressions derived from UnifiedExpressions base parameters. These are also intended to be used as both base parameters and used in merged params. They are different from merged params as they are more recognizable as base params from other standards such as AR52 and SRanipal for example. They also have a conversion map as well in the same .cs file. 

I'm not entirely sure the best approach for this class, or if it even should be separated from the UE file, but I envisioned that as the base UE parameters eventually get segmented out (if the standard continues to grow) that the then child of said parameters would eventually get fed into this class and dynamically iterate the standard whilst keeping backwards compatibility, hence the name.

VRCFaceTracking/SDK/Attributes.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <summary>
+        /// Container of the referenced module GUID.
+        /// </summary>
+        public string ModuleGUID { get; private set; }
+
+        public VRCFTModuleEnsureLoadAfterAttribute(string moduleGUID)
+        {
+            ModuleGUID = moduleGUID;
+        }
+    }
```
regzo2, line 67: Just want to double check and make sure that these attributes are what we expect them to be? VRCFTModuleEnsureLoadAfter will load after a specified module GUID, and VRCFTModuleEnsureLoadPriorAttribute will load before a specified module GUID. This affects the order they appear in the ListBox on the UI, and how they are initialized by default. I believe I have it so that the user order will override this but I can't exactly remember if I allowed that interaction or not.

VRCFaceTracking/SDK/ExtTrackingModule.cs
```diff
@@ -5,13 +5,13 @@ namespace VRCFaceTracking
     public abstract class ExtTrackingModule
     {
         // Should UnifiedLibManager try to initialize this module if it's looking for a module that supports eye or lip.
-        public virtual (bool SupportsEye, bool SupportsLip) Supported => (false, false);
-        
+        public virtual (bool SupportsEye, bool SupportsExpressions) Supported => (false, false);
```
regzo2, line 8: I made the change from SupportsLip to SupportsExpressions early, though notably I accidentally made it plural whereas all other instances of 'expression' are singular, so maybe it should be SupportsExpression? Idk it makes sense either way, but we should be consistent with it (more-so my fault).

VRCFaceTracking/UnifiedConfig.cs
```diff
@@ -0,0 +1,80 @@
...
+            catch (FileNotFoundException)
+            {
+                Logger.Warning("Configuration file not found; creating placeholder 'Config.json' file.");
+                Save();
+            }
+            catch (JsonException)
+            {
+                Logger.Warning("Configuration file has been corrupted, purging file.");
+                File.Delete(unifiedConfigPath);
+            }
```
regzo2, line 55: Specifically for the Config file, I added this early on, however completely purging the file is obviously not ideal if we expect users to be able to tune their data continuously. I am not entirely familiar with Text.Json, but I imagine there is a way to allow it to still propogate data even if the objects are not identical (potentially in the case of future updates to VRCFT) and retain configs thru updates.

VRCFaceTracking/OSC/OSCParams.cs
```diff
@@ -45,13 +45,17 @@ public BaseParam(string name, Type type)
...
 
             public virtual void ResetParam(ConfigParser.Parameter[] newParams)
             {
-                var compatibleParam =
-                    newParams.FirstOrDefault(param => param.name == _paramName && param.input.Type == _paramType);
+                Regex regex = new Regex(@"(?<!(v\d+))(/" + _paramName + @")$|^(" + _paramName + @")$");
+
+                var compatibleParam = newParams.FirstOrDefault(param =>
+                    regex.IsMatch(param.name)
+                    && param.input.Type == _paramType);
```
regzo2, line 57: I wonder if there might be something that could be improved with this regex pattern? It's meant to allow wildcard prefixes and avoids any baked in prefixes (eg. v2/-x-). Anyhow the performance seems fine and doesn't affect the parameter sending performance.

VRCFaceTracking/OSC/OSCParams.cs
```diff
@@ -147,13 +151,15 @@ public override void ResetParam(ConfigParser.Parameter[] newParams)
                 _negativeParam.ResetParam(newParams);
 
                 // Get all parameters starting with this parameter's name, and of type bool
-                var boolParams = newParams.Where(p => p.input.Type == typeof(bool) && p.name.StartsWith(_paramName));
+                Regex regex = new Regex(@"(?<!(v\d+))/" + _paramName + @"\d+$|^" + _paramName + @"\d+$");
+
+                var boolParams = newParams.Where(p => 
+                    p.input.Type == typeof(bool) && regex.IsMatch(p.name));
```
regzo2, line 157: Same as above except for binary parameter parsing.

VRCFaceTracking/Params/Data/UnifiedData.cs
```diff
@@ -0,0 +1,103 @@
...
+namespace VRCFaceTracking.Params
+{
+    /// <summary>
+    /// Struct that represents a single eye.
+    /// </summary>
+    public struct UnifiedSingleEyeData
+    {
+        // Eye data.
+        public Vector2 Gaze;
+        public float PupilDiameter_MM;
```
regzo2, line 14: Needs to be in CM and not MM (to make it clear for module devs and intent of use), there isn't any intrinsic logic to translate this to a 0-1 range. Making this in CM would mean that pupil dilation would definitely be between 0-1.

VRCFaceTracking/Params/Data/UnifiedData.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// <summary>
+    /// Container of information pertaining to a singular Unified Expression shape.
+    /// </summary>
+    public struct UnifiedExpressionShape
+    {
+        /// <summary>
+        /// Value that contains the specified Unified Expression raw value.
+        /// </summary>
+        public float Weight;
+    }
```
regzo2, line 72: I put the weight in a struct early on in the UE builds, and they used to contain all the Mutator data as well, but after moving those out of the Shape structs there only exists Weight now. I'm not entirely sure what else would be contained here in the future but it would only be data that is meant to directly affect the shapes (maybe a Curve shape in the future for example?).

VRCFaceTracking/Params/Data/UnifiedTrackingMutator.cs
```diff
@@ -0,0 +1,168 @@
...
+    public class UnifiedTrackingMutator
+    {
+        public struct Mutation
+        {
+            public string Name;
+            public float Ceil; // The maximum that the parameter reaches.
+            public float Floor; // the minimum that the parameter reaches.
+            //public float SigmoidMult; // How much should this parameter be affected by the sigmoid function. This makes the parameter act more like a toggle.
+            //public float LogitMult; // How much should this parameter be affected by the logit (inverse of sigmoid) function. This makes the parameter act more within the normalized range.
+            public float SmoothnessMult; // How much should this parameter be affected by the smoothing function.
```
regzo2, line 22: Currently the calibrator includes only a ceiling and floor calibration for each parameter, and I included some commented out ones for adjusting the profile of each parameter (one that makes it more toggle like and one that makes it more curve like). Basically the Mutator class was really intended on VRCFT itself allowing users to modify how expressions appear or feel like on a user level, though how this class can be modified by modules isn't yet entirely clear.

VRCFaceTracking/OSC/OSCParams.cs
InconsolableCellist: I'd add a comment explaining the regex at a glance and the purpose of the two parameter naming conventions

VRCFaceTracking/SDK/Attributes.cs
regzo2: Note that EnsureLoadAfter is currently unimplemented, EnsureLoadPrior was what @benaclejames was wanting implemented for specific modules.

VRCFaceTracking/OSC/OSCParams.cs
InconsolableCellist: Also v1/foo/bar would match but v1/bar wouldn't. Is that intentional?

VRCFaceTracking/OSC/OSCParams.cs
```diff
@@ -147,13 +151,15 @@ public override void ResetParam(ConfigParser.Parameter[] newParams)
...
 
                 var paramsToCreate = new Dictionary<string, int>();
                 foreach (var param in boolParams)
                 {
-                    // Cut the parameter name to get the index
-                    if (!int.TryParse(param.name.Substring(_paramName.Length), out var index)) continue;
+                    if (!int.TryParse(String.Concat(param.name.ToArray().Reverse().TakeWhile(char.IsNumber).Reverse()), out var index)) continue;
```
InconsolableCellist, line 162: What's are some good test inputs for this? There might be a simpler way to do it but I want to see what it should work on

regzo2: That part specifically is for parsing out the int from a binary parameter (eg. v2/JawOpen4, v2/JawOpen2, v2/JawOpen1...), it basically counts from end to start and gets the first number it encounters and reverses it back. So v2/JawOpen16 would give int 16 for example. This is to work around the possibility that a loaded param won't strictly contain the base param name (eg. JawOpen) from the regex parsing (which could be like v2/JawOpen).

Basically any sets of binary params would be great to test, even testing out different nested patterns as well, like: 
```
gamer/gunk/v2/JawOpen16
test/v2/JawOpen8
123123123/v2/JawOpen4
w/t/f/v2/JawOpen2
v2/JawOpen1
```

VRCFaceTracking/OSC/OSCParams.cs
regzo2: Yes, though the format of VRCFT parameters will ensure that isn't really an issue (you would have to intentionally name a Unified parameter `foo/bar` to have that happen).

regzo2: I guess an extra `/` should invalidate the entire row

ALVRTrackingInterface/ALVRTrackingInterface.cs
```diff
@@ -0,0 +1,412 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Net;
+using System.Net.Sockets;
+using System.Reflection;
+using System.Threading;
+using VRCFaceTracking;
+using VRCFaceTracking.Params;
+
+namespace ALVRTrackingInterface
```
regzo2, line 11: This applies to all ExtTrackingModules that are included with the VRCFaceTracking repo (ALVR module and SRanipal module respectively):

What do you all think of having the included external modules in separate repos from VRCFaceTracking? I feel that it would help better communicate that this repo is fully encapsulated by the VRCFaceTracking namespace alone, and any external modules are outside of that. It would also help keep commits done to the external modules separate from commits done to the actual VRCFaceTracking namespace? The installer would still include them however.

VRCFaceTracking/SDK/ExtTrackingModule.cs
regzo2: Resolved in https://github.com/benaclejames/VRCFaceTracking/pull/122/commits/aa938e3d5e4e96673962590d56e10897764e2fee

VRCFaceTracking/Params/Data/UnifiedData.cs
regzo2: Actually the diameter is exposed via an explicit implementation (like with eye gaze and the others), it really could just be altered in the merger class.

ALVRTrackingInterface/ALVRTrackingInterface.cs
```diff
@@ -0,0 +1,412 @@
...
+            (
+                LookLeft: expressions[(int)FBExpression.Eyes_Look_Left_R],
+                LookRight: expressions[(int)FBExpression.Eyes_Look_Right_R],
+                LookUp: expressions[(int)FBExpression.Eyes_Look_Up_R],
+                LookDown: expressions[(int)FBExpression.Eyes_Look_Down_R]
+            );
+
+            // Eye dilation code, automated process maybe?
+            eye.Left.PupilDiameter_MM  = 0.0035f;
+            eye.Right.PupilDiameter_MM = 0.0035f;
```
Adjerry91, line 277: EyesDilation on legacy parameters is being driven to 0 with input from ALVR, users must disable eye dilation to go back to default. I believe the mutators zeroing out the pupil diameter with no input. Previous QuestPro controls would just hold default value of 0.73 on the float value.

benaclejames:
image
Not sure what's going on here. This error is thrown when using vrcft on a fresh machine with an empty appdata folder

regzo2:
> image Not sure what's going on here. This error is thrown when using vrcft on a fresh machine with an empty appdata folder

SRanipal SDK libs need to be in the ModuleLibs folder, though I was wondering if embedding them in the dll would be more ideal?

ALVRTrackingInterface/ALVRTrackingInterface.cs
regzo2: The .73 value you're talking about was controlling the old EyesDilation parameter directly (which just represented the value between 0-1), currently the value that is expected is the exact MM value from an interface, which then on runtime is used to create the old EyesDilation ranges. 

I can just have it so that this interface sets the dilation ranges from 0 - 10, and then the pupil diameter can be set to whatever it needs to be (which we would expect to be between 5 - 6 MM for normal pupil diameter ranges)

regzo2: See if https://github.com/benaclejames/VRCFaceTracking/pull/122/commits/74f3b62885b1ce48787ac666778226e703c7801b resolves your issue!

PinballsWizard:
> There are a few things that we need to make sure are retained as the conflicts get merged that are already a part of main, and might not necessarily show up as conflicts b/c of the file restructure (as of Nov 5, 2022):

With regards to points 1 and 3:
1. [fc76bf9](https://github.com/benaclejames/VRCFaceTracking/commit/fc76bf9dc2b4492664b013cd82de860340f8d68b) - disable-args
3. [0865da9](https://github.com/benaclejames/VRCFaceTracking/commit/0865da9e20c3ff1748e4368c78feaf15c75beb5a) - possibly these changes to UnifiedLibManager (from disable-args changes)?

Both the initial changes for disable-args and the related changes were recreated/present in this PR in [2135e8c](https://github.com/benaclejames/VRCFaceTracking/pull/122/commits/2135e8c27d094b66c5efe3a626f0b544f6b014c1)

I'm sure it will need to be touched again for [refactor/args](https://github.com/benaclejames/VRCFaceTracking/tree/refactor/args) later as well

korejan:
Please rename the ALVR module to ALXR, it is the ALXR client that the module recieves the data from, not the alvr server. The alxr client supports both (Air)Link users without the ALVR server and ALVR users.

It will make less sense soon to call it ALVR as I'm currently working on adding more facial & eye tracking OpenXR extensions (for example `XR_HTC_facial_tracking` is done) to ALXR and updaing the vrcft module to support them. This is for both wired and standalone headsets. I am also planning to make new option for the module to use the alxr-engine DLL directly.

Adjerry91:
Loading VRCFacetracking it does not load the module on start in the UI you have to click the reload modules. 
image

regzo2:
> Please rename the ALVR module to ALXR, it is the ALXR client that the module recieves data from, not the alvr server. For Quest Pro, the alxr client supports both (Air)Link users without the ALVR server and ALVR-server users.
> 
> It will make less sense soon to call it ALVR as I'm currently working on adding more facial & eye tracking OpenXR extensions (for example `XR_HTC_facial_tracking` is done) to ALXR and updaing the vrcft module to support them. This is for both wired and standalone headsets. I am also planning to make a new option for the module to use the alxr-engine DLL directly (for OpenXR PC runtimes like Link).

I personally think having the ALVR/ALXR external tracking module in it's own repo would be best so that you can make changes needed to the module (and I'm trying to get opinions from the other contributors on moving the other external tracking modules that are in this repo into their own repos as well). If you want to make a dedicated repo for the ALVR/ALXR module then feel free to do so!

regzo2:
> Loading VRCFacetracking it does not load the module on start in the UI you have to click the reload modules. image

This is mostly a result of the main thread and UI threads not entirely initializing synchronously, and also having the UI listbox fetching module data on init. A lot of the initialization processes such as the module listbox try loading data on UI init, which is reliant on the main being fully propogated with data. We could look into having a finalized init on main finalizing or hooking into WPF events to propogate data to the UI to circumvent this, though if we plan on moving to a newer UI framework then it might be best to wait for that (and i'm not sure of the release window for this current build).

korejan:
> > Please rename the ALVR module to ALXR, it is the ALXR client that the module recieves data from, not the alvr server. For Quest Pro, the alxr client supports both (Air)Link users without the ALVR server and ALVR-server users.
> > It will make less sense soon to call it ALVR as I'm currently working on adding more facial & eye tracking OpenXR extensions (for example `XR_HTC_facial_tracking` is done) to ALXR and updaing the vrcft module to support them. This is for both wired and standalone headsets. I am also planning to make a new option for the module to use the alxr-engine DLL directly (for OpenXR PC runtimes like Link).
> 
> I personally think having the ALVR/ALXR external tracking module in it's own repo would be best so that you can make changes needed to the module (and I'm trying to get opinions from the other contributors on moving the other external tracking modules that are in this repo into their own repos as well). If you want to make a dedicated repo for the ALVR/ALXR module then feel free to do so!

I want to apologize for being rude, sorry for being confrontational.

I'm totally fine with making a dedicated repo. It would have been what I would do but I wasn't originally involved on the vrcft module side of it until recently.

I have a fork of the quest-pro branch and have branches off of it with my work so I'm regularly rebasing them.

Let me know how/when you want do this and thank for updating that module for the unified expression work.

regzo2:
> > > Please rename the ALVR module to ALXR, it is the ALXR client that the module recieves data from, not the alvr server. For Quest Pro, the alxr client supports both (Air)Link users without the ALVR server and ALVR-server users.
> > > It will make less sense soon to call it ALVR as I'm currently working on adding more facial & eye tracking OpenXR extensions (for example `XR_HTC_facial_tracking` is done) to ALXR and updaing the vrcft module to support them. This is for both wired and standalone headsets. I am also planning to make a new option for the module to use the alxr-engine DLL directly (for OpenXR PC runtimes like Link).
> > 
> > 
> > I personally think having the ALVR/ALXR external tracking module in it's own repo would be best so that you can make changes needed to the module (and I'm trying to get opinions from the other contributors on moving the other external tracking modules that are in this repo into their own repos as well). If you want to make a dedicated repo for the ALVR/ALXR module then feel free to do so!
> 
> I want to apologize for being rude, sorry for being confrontational.
> 
> I'm totally fine with making a dedicated repo. It would have been what I would do but I wasn't originally involved on the vrcft module side of it until recently.
> 
> I have a fork of the quest-pro branch and have branches off of it with my work so I'm regularly rebasing them.
> 
> Let me know how/when you want do this and thank you for updating the module for the unified expression work.

Here is an update: We have talked about what our plans are as far as rebasing the modules. Currently we are interested in working towards a rebase of all the modules into their own repositories, but nothing is completely concrete in when or how we are going to do so. We are planning on making this a focus at the very least after the next VRCFT release.

PR #125, opened 2023-03-25
Optimize OSC Implementation

benaclejames:
This PR utilizes a [rust-made .dll](https://github.com/benaclejames/SimpleRustOSC) to handle the bulk of OSC serialization/deserialization. It helps a little with optimization but also clears up our internal codebase a bit!


VRCFaceTracking/MainStandalone.cs
```diff
@@ -60,6 +51,7 @@ public static void Teardown()
         
         public static void Initialize()
         {
+            var test2 = BitConverter.GetBytes(69.42);
```
regzo2, line 54: Unused variable (nice)

VRCFaceTracking/MainStandalone.cs
```diff
@@ -95,23 +87,9 @@ public static void Initialize()
...
-                    .Where(param => param.Relevant);
-                _relevantParams_v2 = UnifiedTracking.AllParameters_v2.SelectMany(p => p.GetBase())
-                    .Where(param => param.Relevant);
-
-                _relevantParams = _relevantParams_v1.Concat(_relevantParams_v2);
-
-                Logger.Msg("Config file parsed successfully! " + _relevantParams.Count() + " parameters loaded.");
-                
-                if (_relevantParams_v1.Count() > 0)
-                    Logger.Warning(_relevantParams_v1.Count() + " legacy parameters loaded. These are undocumented and outdated parameters.");
```
regzo2, line 114: Noticed that the loaded event doesn't have a way to distinguish if a v1 parameter was loaded, and the legacy message would always load even when the newer v2 parameters get loaded. Could have the lambda return something to indicate that the relevant v1 params got loaded (could be a lowest version loaded int or flags to indicate parameter versions loaded), perhaps they could be all contained in a ConfigInfo struct or something and sent as the lambda parameter?

Unless you are thinking of not having the legacy message at all?

VRCFaceTracking/ConfigParser.cs
```diff
@@ -62,10 +62,11 @@ public static void ParseNewAvatar(string newId)
             Logger.Msg("Parsing config file for avatar: " + avatarConfig.name);
             var parameters = avatarConfig.parameters.Where(param => param.input != null).ToArray();
 
+            int count = 0;
             foreach (var parameter in UnifiedTracking.AllParameters_v2.Concat(UnifiedTracking.AllParameters_v1).ToArray())
-                parameter.ResetParam(parameters);
+                count += parameter.ResetParam(parameters);
```
regzo2, line 67: Where we could detect loaded AllParameter versions as well for the ConfigLoaded action.

VRCFaceTracking/OSC/OSCMessage.cs
```diff
@@ -1,164 +1,115 @@
...
+    public struct OscMessageMeta {
+        [MarshalAs(UnmanagedType.LPStr)]
+        public string Address;
+        
+        public OscValueType Type;
+        
+        public OscValue Value;
+    }
+    
+    public static class RustLib
```
regzo2, line 37: Might make sense to specify the class a bit more (RustOSCLib perhaps?), though it is under the OSC namespace.

regzo2 (changes requested):
Found some things that might be worth looking at. I'll leave it up to you if you wanna make those changes! I haven't look at your Rust OSC lib but these are just addressing the interfacing of it (what you're probably looking for). Everything else looks pretty good from looking at the other changes.

Also have yet to test it but just some observations of the code.

regzo2 (changes requested):
Just one final thing related to the legacy msg. Everything under the hood works exactly as expected!

VRCFaceTracking/MainStandalone.cs
```diff
@@ -87,9 +86,9 @@ public static void Initialize()
             if (!bindResults.senderSuccess)
                 Logger.Error("Socket failed to bind to sender port, please ensure it's not already in use by another program or specify a different one instead.");
 
-            ConfigParser.OnConfigLoaded += (count) =>
+            ConfigParser.OnConfigLoaded += relevantParams =>
             {
-                    Logger.Warning(count + " legacy parameters loaded. These are undocumented and outdated parameters.");
+                Logger.Warning(relevantParams.Count(p => p.Relevant) + " legacy parameters loaded. These are undocumented and outdated parameters.");
             };
```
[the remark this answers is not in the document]
benaclejames: image
Quick 'n dirty, for now..

VRCFaceTracking/OSC/OSCParams.cs
```diff
@@ -9,12 +9,13 @@ namespace VRCFaceTracking.OSC
     public class OSCParams
     {
         private const string DefaultPrefix = "/avatar/parameters/";
+        private const string CurrentVersionPrefix = "v2/";
```
regzo2, line 12: Kinda makes me think we should include some sort of versioning within the IParameter side of the parameter eventually, and use OSCParams or another wrapper that enforces it (like with the base parameters getting added to AllParameters for eg.). 

I think for now this works pretty good and we can always revisit it ofc!

VRCFaceTracking/OSC/OSCParams.cs
```diff
@@ -201,10 +190,11 @@ public override IParameter[] ResetParam(ConfigParser.Parameter[] newParams)
                 return null;
             }
 
-            protected BinaryBaseParameter(string paramName, Func<UnifiedTrackingData, float> getValueFunc) : base(paramName, getValueFunc)
+            protected BinaryBaseParameter(string paramName, Func<UnifiedTrackingData, float> getValueFunc)
             {
                 _paramName = paramName;
-                _negativeParam = new BaseParam<bool>(paramName + "Negative");
+                _getValueFunc = getValueFunc;
+                _negativeParam = new BaseParam<bool>(paramName + "Negative", data => getValueFunc.Invoke(data) < 0);
```
regzo2, line 197: Also another thing we could consider doing in another PR in the future: if there isn't a negative parameter parsed then it won't return the negative values on output (there might be some cases where an avatar creator wants to use the positive part of a combined param and exclude the negative bool to save a little space).

regzo2 (approved):
Looks good! Everything is working as expected and the comments are just for things we might want to do in the future (if at all).

Also for your Rust lib you could bring back the [DependencyManager](https://github.com/benaclejames/VRCFaceTracking/blob/3ea923d89c8a7383709558ebcdf707babe1ef93f/VRCFaceTracking/DependencyManager.cs), I had removed it from the SRanipal rejection from the code but it would make sense for this (or even embedding the compiled binary into the EXE somehow like how ILMerge or Fody.Costura can be used to achieve that).

VRCFaceTracking/OSC/OSCParams.cs
benaclejames: This is already the case! Binary bits check to ensure the negative parameter is relevant before sending. If not and the value would be negative, they just return false
https://github.com/benaclejames/VRCFaceTracking/pull/125/files/5fa1874372286802b9c2d04cf6a0c7f6f3f9eca3..fe8709c89c71dcd946767fcb05435204076c54fe#diff-5ed98eabd76f774c20ea2dc83b3436d6ff97b8b5df8f0ea0d7b07facd30a0d27R115-R117

benaclejames: Though maybe you mean allowing it to return a value regardless of whether or not the negative parameter is relevant or not (might be misunderstanding you here). Though as you said, can always be fixed in the future :)

PR #129, opened 2023-04-21
Optimize localized

TheBug233:


TheBug233:
missing a colon..resubmit later

PR #130, opened 2023-04-21
Optimize localization code and add Simplified Chinese language

TheBug233:


benaclejames (approved):
Looks good, and wow that was FAST! Thank you so much :)

PR #131, opened 2023-04-23
Plural Be Gone! (Removed all plural naming in v2 combined parameters)

regzo2:
Pretty self explanatory, simply removes all plural naming in combined parameters for v2.
Example for BrowDown expressions:
``` 
BrowDown -> BrowDownRight
         -> BrowDownLeft
 ```
 This forces all parameters to not only be consistent in the naming convention, but also to follow a naming convention now. I hope this change reduces ambiguity when picking combined parameters in v2 parameters!
 
 I do not believe this will affect many people, since most people are still currently on legacy parameters


benaclejames (approved):
Looks good! Provided params are documented as to a lack of plural resulting in a combined parameter, I see no reason why there would be confusion

PR #132, opened 2023-04-23
v2 Parameter Adjustments and Fixes

regzo2:
Whoops! We missed these when converting to the new convention. CheekPuffSuck parameters are now consistent with all parameters now. Will lay out the conventions of parameter naming in the future!

I also cleaned up the ToPitch and ToYaw functions and cleaned up the Native parameters as much as I could.

EyesClosedAmount also now properly searches for eye lid parameters instead of gaze parameters


benaclejames (approved):
yip

PR #133, opened 2023-04-24
Add ExpressionTrackingActive

PinballsWizard:
As part of the larger Lip -> Expression update

Seem to have been missed in the overall update, I know the special case handling the parameter changing in game making it back to VRCFT is commented out for now but that was the only place I could find it actually in the code, so this should be what's needed to have it actually function.

Of note also ExpressionTrackingActive would then also need to be on [the wiki docs](https://github.com/benaclejames/VRCFaceTracking/wiki/Parameters).

### DIFF
--- a/VRCFaceTracking.Core/Params/Expressions/UnifiedExpressionsMerger.cs
+++ b/VRCFaceTracking.Core/Params/Expressions/UnifiedExpressionsMerger.cs
@@ -281,6 +281,7 @@ public static class UnifiedExpressionsMerger
         #endregion
 
         new ConditionalBoolParameter(exp => (UnifiedLibManager.EyeStatus == ModuleState.Active, UnifiedLibManager.EyeStatus != ModuleState.Uninitialized), "EyeTrackingActive"),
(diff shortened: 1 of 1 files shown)


regzo2 (approved):
Looks good to me!

benaclejames:
Hey, sorry but why are we adding this? Don't we have LipTrackingActive with the exact same functionality? Am I missing something here? 😆

PinballsWizard:
Might be better left to @regzo2 for further rationale, but as I understand it everywhere else where Lip was used previously has been moved to Expression with the UE update (such as the launch args), so in this case LipTrackingActive should act as an alias/backwards compatibility for ExpressionTrackingActive

regzo2:
> Hey, sorry but why are we adding this? Don't we have LipTrackingActive with the exact same functionality? Am I missing something here? 😆

This is just to align with the rest of the changes done with VRCFT. Expressions is a soft replacement for Lips, and it includes even more than just lips now including 'Eye' expressions like eye widen, squinting, etc.

PR #137, opened 2023-04-25
[v5.0.0] Parameters/Negative Binary fix + Config parsing

regzo2:
This PR fixes the dreaded TongueSteps bug by reintegrating negativeParams into ResetParam exclusively. 

I also slightly adjusted the config parser to be able to propogate OSC values again when the avatar ID's are the same (it skips reading the OSC directory and just sends to the now statically saved AvatarConfigSpec).

Let me know if these are good so @Adjerry91 can be happy again, thx


benaclejames:
Hmm, why are we re-propegating when (provided the avatar is the same), the parameters wouldn't have changed?

regzo2:
Essentially the propogation will always ensure the value gets set properly (in the case of specific parameters like TongueSteps, it will ensure that the 'default' value of -1.0 gets set properly). It might be good practice to do that for all avatars.

PR #138, opened 2023-04-25
Update MainStandalone.cs

nkotech:
Warning instead of error for legacy parameters


benaclejames (approved):
*huh* thought I did this already, guess I forgot to actually commit it. Thanks!

PR #146, opened 2023-05-09
Added references to the new docs

nkotech:
Updated the links that have docs.vrcft.io counterparts and updated the picture


benaclejames (approved):
epic

PR #156, opened 2023-07-17
Fix use of vcruntime140.dll

zarik5:
Use absolute path instead of local path for fetching `vcruntime140.dll`.


zarik5:
#157 does the same thing as this PR but better. Closing.

PR #157, opened 2023-07-22
Update VRCFaceTracking.Core.csproj

frg2089:
Use environment variables instead of this relative path.


benaclejames:
Fixed using appinstaller method in https://github.com/benaclejames/VRCFaceTracking/commit/e2ed80aca995df4c20e8b135e0157cbd5186eccd

PR #159, opened 2023-07-29
Optimize and update the newly added localized content

TheBug233:
Optimize and update the newly added localized content. I made English and Simplified Chinese. Other languages can add new translation content by comparing the new value in English


benaclejames:
Hey, we've discussed this internally and we'd quite like to keep the log messages in English so we're actually able to debug them when they're reported to us. We would, however, love to accept any localization for the UI (button text, settings text, etc)

TheBug233:
ok, i'll resubmit later

Issue #162, opened 2023-08-04
The Install module button does nothing


orcinus:
Reproduction steps:
1) click "Install some from here" from the "No Modules Installed" notification
2) click "Install" button for SRanipalTrackingModule
3) nothing happens

Expected result:
Module gets downloaded and installed.

Version: 5.1.1.0

Platform: Windows 10 22H2

orcinus:
Log output:

```
[UnifiedLibManager]` Information: Starting initialization tracking
[UnifiedLibManager] Information: Tearing down all modules...
[UnifiedLibManager] Debug: Initializing requested runtimes...
[UnifiedLibManager] Information: Initializing runtimes...
[UnifiedLibManager] Warning: No modules loaded.
[VRCFaceTracking.Core.Services.ModuleDataService] Debug: Rating for 74f18332-1bcf-4b55-832d-9c91069afa31 was not found
[UnifiedLibManager] Information: Tearing down all modules...
[VRCFaceTracking.Core.Services.ModuleInstaller] Debug: Uninstalling module 74f18332-1bcf-4b55-832d-9c91069afa31
[VRCFaceTracking.Core.Services.ModuleInstaller] Warning: Module 74f18332-1bcf-4b55-832d-9c91069afa31 could not be found where it was expected in C:\Users\Ante Vukorepa\AppData\Roaming\VRCFaceTracking\CustomLibs\74f18332-1bcf-4b55-832d-9c91069afa31
[App] Error: Unhandled exception
[App] Critical: Stacktrace:    at System.IO.FileSystem.MoveDirectory(String sourceFullPath, String destFullPath, Boolean isCaseSensitiveRename)
   at System.IO.FileSystem.MoveDirectory(String sourceFullPath, String destFullPath)
   at VRCFaceTracking.Core.Services.ModuleInstaller.InstallRemoteModule(TrackingModuleMetadata moduleMetadata)
   at VRCFaceTracking.Views.ModuleRegistryDetailControl.Install_Click(Object sender, RoutedEventArgs e)
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
   at Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.<>c__DisplayClass2_0.<Post>b__0()
[App] Critical: Inner exception: (null)
[App] Critical: Message: Source and destination path must have identical roots. Move will not work across volumes.
[UnifiedLibManager] Information: Tearing down all modules...
```

[...]

PR #164, opened 2023-08-04
Make module folder move work across volumes

orcinus:
Fix for issue #162 

Directory.Move does not work across volumes, meaning having your Temp folder not on the same drive as your User folder will cause module installation to fail.

This uses a non-recursive folder-copy / delete appropriated for SO instead of Directory.Move.


orcinus:
Sorry about the whitespace crap, that was the VS linter's doing.

benaclejames:
Sorry for the delay on this, been super busy implementing the new oscquery spec as per vrchat's open beta. I actually made & committed a less-verbose fix for this on the day but forgot to actually push it. Thanks for your PR and sorry I didn't see it sooner

orcinus:
No worries, important thing is it's fixed :)

orcinus:
This still hasn't been integrated into a release?

PR #190, opened 2024-01-25
change m/s to msgs/sec in OSC section in main page

kusomaigo:
because way too many people have been thinking it is ms of latency


nkotech (approved):
Yeah that's better

nkotech (approved):
attempt 2 to approve

benaclejames (approved):
Never figured people would be this confused. There's literally zero way to have any network latency if the data isn't leaving your computer. It appears my assumption that this was common knowledge was false.

Then again, we've had people come into the discord asking for help getting their quest 2's non-existent eye tracking working, so I should've seen it coming

PR #195, opened 2024-02-15
Parameter name consistency fix

RantiMess:
Changed 
v2/MouthTightenStretch/Left/Right to
v2/MouthTightenerStretch/Left/Right for consistency with existing v2/MouthTightener parameter

Nice catch jerry


benaclejames (approved):
Tysm!

PR #215, opened 2024-06-18
Fix normal osc avatar change and force relevancy

### LINKED ISSUE #212: forceRelevant does

hazre:
in v5.2.3 both `/vrcft/settings/forceRelevant` and `/avatar/change` parameters stopped working when using normal osc without OscQuery.

This PR reverts to how it behaved in v5.1.1 while still keeping the new OscQuery stuff.

closes #212


hazre:
@benaclejames sorry for the mess, I cleaned up the commits, so there is no more whitespace changes. It should be pretty self explanatory now.

benaclejames (approved):

PR #230, opened 2024-10-01
Prevent newer (unreleased) Local Modules being overwritten by Registry.

regzo2:
Simple as the title suggests. Changes registry activation behavior to prevent local modules being overwritten in the case that they have a higher version number, as opposed to a different version regardless of higher versioning or not.

### DIFF
--- a/VRCFaceTracking/Services/ActivationService.cs
+++ b/VRCFaceTracking/Services/ActivationService.cs
@@ -119,20 +119,14 @@ public class ActivationService : IActivationService
         var outdatedModules = remoteModules.Where(rm => localModules.Any(lm =>
         {
             if (rm.ModuleId != lm.ModuleId) 
-            {
                 return false;
-            }
 
-            var remoteVersSplit = rm.Version.Split(".");
-            var localVersSplit = lm.Version.Split(".");
+            var remoteVersion = new Version(rm.Version);
+            var localVersion = new Version(lm.Version);
 
-            for (int i = 0; i < localVersSplit.Length; i++)
-            {
-                if (i > remoteVersSplit.Length) 
-                    return false;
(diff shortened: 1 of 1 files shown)


benaclejames (changes requested):
Why not just use System.Version?

regzo2:
> Why not just use System.Version?

Version has a limit on the number of segments (4 specifically: major, minor, build, rev), unless that is what you want VRCFT to enforce that schema?

benaclejames:
I think that should be enough, especially since all modules that currently exist in the registry conform to that schema. Should probably document this behaviour though.

regzo2:
> I think that should be enough, especially since all modules that currently exist in the registry conform to that schema. Should probably document this behaviour though.

Sounds good 👍

VRCFaceTracking/Services/ActivationService.cs
```diff
@@ -116,7 +116,20 @@ private async Task StartupAsync()
...
+                return false;
+
+            var remoteVersion = new Version(rm.Version);
+            var localVersion = new Version(lm.Version);
+
+            var result = remoteVersion.CompareTo(localVersion);
+            if (result > 0) // remote module version is higher than registry
+                return true;
+
+            return false;
```
benaclejames, line 131: Could this be simplified to `return remoteVersion.CompareTo(localVersion) > 0;`?

regzo2 (commented):
Adjusted to use Version for the comparison

VRCFaceTracking/Services/ActivationService.cs
regzo2: Sure

benaclejames (approved):

PR #233, opened 2024-10-07
Fixed crash on invalid ip and or port change

ThatGuyThimo:
When changing the ip address or port number to an invalid value on the next restart the app would crash do to the loaded invalid value'(s),

This pull requests checks if: the ip address is not blank or white space, the ip is an valid ip address, it is a resolvable hostname.
It also checks if the ports are not smaller than 1.

Let me know if the code needs any modifications.

### DIFF
--- a/VRCFaceTracking.Core/Models/OscTarget.cs
+++ b/VRCFaceTracking.Core/Models/OscTarget.cs
@@ -1,11 +1,11 @@
 ﻿using CommunityToolkit.Mvvm.ComponentModel;
 using VRCFaceTracking.Core.Contracts;
 using VRCFaceTracking.Core.Contracts.Services;
-using System.Net;
+using VRCFaceTracking.Core.Validation;
 
 namespace VRCFaceTracking.Core.Models;
 
-public partial class OscTarget : ObservableObject, IOscTarget
+public partial class OscTarget : ObservableValidator, IOscTarget
 {
     [ObservableProperty] private bool _isConnected;
 
@@ -21,39 +21,11 @@ public partial class OscTarget : ObservableObject, IOscTarget
     [NotifyPropertyChangedFor(nameof(InPort))] [NotifyPropertyChangedFor(nameof(OutPort))]
 
     [property: SavedSetting("OSCAddress", "127.0.0.1")]
+    [ValidIpAddress]
     private string _destinationAddress;
-
-    private const string DefaultAddress = "127.0.0.1";
-    private bool _isReverting = false;
 
     public OscTarget(ILocalSettingsService localSettingsService)
     {
         PropertyChanged += (_, _) => localSettingsService.Save(this);
     }
-
-    /// Called after the DestinationAddress property has changed.
-    /// It validates the new address and reverts it to the default if invalid.
-    partial void OnDestinationAddressChanged(string oldValue, string newValue)
-    {
-        // Prevent re-entrancy if we're already reverting
-        if (_isReverting)
-        {
-            return;
-        }
-
-        _isReverting = true;
-
-        // Check if the new address is valid and revert if invalid
-        if (!IsValidAddress(newValue))
-        {
-            DestinationAddress = DefaultAddress;
-        }
-
-        _isReverting = false;
-    }
-
-    /// Validates whether the provided address is a valid IP address or resolvable hostname.
-    private static bool IsValidAddress(string address) =>
-        !string.IsNullOrWhiteSpace(address) &&
-        IPAddress.TryParse(address, out _);
 }
--- a/VRCFaceTracking.Core/Models/OscTarget.cs
+++ b/VRCFaceTracking.Core/Models/OscTarget.cs
@@ -1,6 +1,7 @@
 ﻿using CommunityToolkit.Mvvm.ComponentModel;
 using VRCFaceTracking.Core.Contracts;
 using VRCFaceTracking.Core.Contracts.Services;
+using System.Net;
 
 namespace VRCFaceTracking.Core.Models;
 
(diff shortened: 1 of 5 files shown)


VRCFaceTracking.Core/Models/OscTarget.cs
```diff
@@ -9,18 +10,94 @@ public partial class OscTarget : ObservableObject, IOscTarget
...
+        try
+        {
+            var hostEntry = Dns.GetHostEntry(address);
+            return true;
+        }
+        catch
+        {
+            // DNS resolution failed
+            return false;
+        }
```
benaclejames, line 101: Could this be simplified as 
```csharp
return Uri.CheckHostName(address) != UriHostNameType.Unknown
```
That way you could avoid the exception handling, as well as resolve for hostnames ***and*** IP addresses all in a single line.

VRCFaceTracking.Core/Models/OscTarget.cs
```diff
@@ -9,18 +10,94 @@ public partial class OscTarget : ObservableObject, IOscTarget
...
+    }
```
benaclejames, line 102: If you do elect to use `Uri.CheckHostName`, this whole function could probably be simplified to
```csharp
private static bool IsValidAddress(string address) =>
    !string.IsNullOrWhiteSpace(address) && 
    (IPAddress.TryParse(address, out _) || 
    Uri.CheckHostName(address) != UriHostNameType.Unknown);
```

VRCFaceTracking.Core/Models/OscTarget.cs
```diff
@@ -9,18 +10,94 @@ public partial class OscTarget : ObservableObject, IOscTarget
...
+
+    /// Called after the OutPort property has changed.
+    /// It checks if the OutPort is valid and reverts to the default if it's invalid (less than 1).
+    partial void OnOutPortChanged(int oldValue, int newValue)
+    {
+        if (newValue < 1)
+        {
+            OutPort = DefaultOutPort;
+        }
+    }
```
benaclejames, line 75: I worry this will cause an infinite loop. Also, such functionality could be achieved by editing values in the ui xaml here https://github.com/benaclejames/VRCFaceTracking/blob/3480ca415c08d0d1e7cdb2a4a8e298beca87663c/VRCFaceTracking/Views/SettingsPage.xaml#L64
and here
https://github.com/benaclejames/VRCFaceTracking/blob/3480ca415c08d0d1e7cdb2a4a8e298beca87663c/VRCFaceTracking/Views/SettingsPage.xaml#L78

benaclejames: Also, not sure you want to be using these operators. They both make it so that the port can only be set if the port is **less than** 1..

benaclejames (changes requested):

VRCFaceTracking.Core/Models/OscTarget.cs
ThatGuyThimo: yeah i will use that instead than the port check does not have to exist,

As for the IP verification i might change the way i approach that since host names are disallowed as it will throw an exception in 
https://github.com/benaclejames/VRCFaceTracking/blob/25f7e9548505978be648e8ecf3a507afc14103c3/VRCFaceTracking.Core/Services/OscRecvService.cs#L49

It might also be smart to do this in 
https://github.com/benaclejames/VRCFaceTracking/blob/25f7e9548505978be648e8ecf3a507afc14103c3/VRCFaceTracking.Core/Services/OscRecvService.cs#L30C1-L51C6

As there is already some handling in there but it seems to actually not handle empty strings but does handle NULL ref

ThatGuyThimo: Although this would still cause the application to crash on restart since it doesn't reset the value.

VRCFaceTracking.Core/Models/OscTarget.cs
```diff
@@ -9,18 +10,50 @@ public partial class OscTarget : ObservableObject, IOscTarget
...
     public OscTarget(ILocalSettingsService localSettingsService)
     {
         PropertyChanged += (_, _) => localSettingsService.Save(this);
     }
-}
\ No newline at end of file
+
+    /// Called after the DestinationAddress property has changed.
+    /// It validates the new address and reverts it to the default if invalid.
+    partial void OnDestinationAddressChanged(string oldValue, string newValue)
```
benaclejames, line 36: Might be an idea to use `NotifyDataErrorInfo` for this instead of just reverting the val
https://learn.microsoft.com/en-us/dotnet/communitytoolkit/mvvm/generators/observableproperty#requesting-property-validation

benaclejames: Might even be able to create a custom validationattribute to validate it's a valid IP and be rid of that whole method

ThatGuyThimo: Will look into it. Seems promising would also like to not have an validating function in the class if its not necessary.

ThatGuyThimo: > Might be an idea to use `NotifyDataErrorInfo` for this instead of just reverting the val https://learn.microsoft.com/en-us/dotnet/communitytoolkit/mvvm/generators/observableproperty#requesting-property-validation

this is not directly possible since NotifyDataErrorInfo is an inheritance from ObservableProperty while the OSCTarget class is an child of ObservableObject which does not support it.

benaclejames: Wat? The parent of OscTarget has no relevance. You should just be able to append the attribute below the existing ObservableProperty attribute of DestinationAddress

ThatGuyThimo:
Maybe Remove the IsNullOrEmpty check in OscRecvService and OscSendService since that is now being handled by the validator.
https://github.com/ThatGuyThimo/VRCFaceTracking/blob/37b39c8e1ffcb444e91e8e04bd68107cca93d85a/VRCFaceTracking.Core/Services/OscRecvService.cs#L45C1-L49C14

VRCFaceTracking.Core/Models/OscTarget.cs
ThatGuyThimo: Idk man it didn't allow me to do it this way, i just ended up doing your second suggestion which is a more elegant solution anyway in my opinion.

PR #242, opened 2024-11-18
Adjust Binary Parameter values to be rounded by midpoint.

### LINKED ISSUE #241: Unexpected Binary Parameters converting
I was expecting the conversion to use rounding, but instead, it round down.
I encountered this issue across modules for Quest VD, Pico, and VIVE Facial Tracker.
VRCFT version: I tested in 5.1.1.0 and 5.2.3.0

For 1-bit quantized parameters, such as converting:

JawOpen [float] to JawOpen1 [bool]

I expected values below 0.5 to convert to false, and values 0.5 and above to convert to true, like this:
- JawOpen = 0 → JawOpen1 = false
- JawOpen = 0.6 → JawOpen1 = true
- JawOpen = 1 → JawOpen1 = true

However, in reality, JawOpen1 only turns true if JawOpen is exactly 1 or higher:
- JawOpen = 0 → JawOpen1 = false
- JawOpen = 0.6 → JawOpen1 = false
- JawOpen = 1 → JawOpen1 = true

The same issue applies to higher bit quantization, but it’s most noticeable with 1-bit, where it can result in up to a 99% discrepancy.

This causes extreme cases where a value of 0.99 is still considered false.

For example, on the Pico 4 Pro, the TongueOut parameter only goes up to 0.99, meaning TongueOut1 from VRCFT is always false.

regzo2:
Adjusts a probable oversight where Binary Parameter bits would always round down to now be based on the midpoint of the parameter bit.

Changed function:
```c#
// BinaryBaseParameter.cs
var adjustedValue = Math.Abs(value);
```
To:
```c#
// BinaryBaseParameter.cs
var adjustedValue = Math.Abs(value) + (1f / _maxPossibleBinaryInt);
```

Resolves #241

### DIFF
--- a/VRCFaceTracking.Core/OSC/DataTypes/BinaryBaseParameter.cs
+++ b/VRCFaceTracking.Core/OSC/DataTypes/BinaryBaseParameter.cs
@@ -21,7 +21,7 @@ public class BinaryBaseParameter : Parameter
         if (!_negativeParam.Relevant &&
             value < 0) // If the negative parameter isn't set, cut the negative values
             return false;
-        var adjustedValue = Math.Abs(value);
+        var adjustedValue = Math.Abs(value) + (1f / _maxPossibleBinaryInt);
         var bigValue = (int)(adjustedValue * (_maxPossibleBinaryInt - 1));
         return ((bigValue >> binaryIndex) & 1) == 1;
     }


benaclejames (approved):
God it's been so long since I last looked at this code
If you've tested it and it works, good enough for me :p

PR #248, opened 2025-01-08
Add head tracking data support

kusomaigo:
Gives VRCFT the ability to support desktop "VTubing" using specially set-up avatars with 6 new internal parameters

```
v2/Head/Yaw
v2/Head/Pitch
v2/Head/Roll
v2/Head/PosX
v2/Head/PosY
v2/Head/PosZ
```

Any "desktop" face-tracking application already provides at least rotation data, and existing modules can easily be adapted to provide these data points. 

Example videos: 

https://github.com/user-attachments/assets/61fab05c-f269-427f-94e8-7d72b0caf3ea

---

https://github.com/user-attachments/assets/314e3d36-930e-4a9e-aa67-2e4197f5de39

### DIFF
--- a/VRCFaceTracking.Core/AvatarConfigFileParser.cs
+++ b/VRCFaceTracking.Core/AvatarConfigFileParser.cs
@@ -68,7 +68,7 @@ public class AvatarConfigParser
         ParameterSenderService.Clear();
         var parameters = avatarConfig.parameters.Where(param => param.input != null).ToArray<IParameterDefinition>();
 
-        foreach (var parameter in UnifiedTracking.AllParameters_v2.Concat(UnifiedTracking.AllParameters_v1).ToArray())
+        foreach (var parameter in UnifiedTracking.AllParameters_v2.Concat(UnifiedTracking.AllParameters_v1).Concat(UnifiedTracking.HeadParameters).ToArray())
         {
             paramList.AddRange(parameter.ResetParam(parameters));
         }
--- a/VRCFaceTracking.Core/AvatarConfigFileParser.cs
+++ b/VRCFaceTracking.Core/AvatarConfigFileParser.cs
@@ -68,7 +68,7 @@ public class AvatarConfigParser
         ParameterSenderService.Clear();
         var parameters = avatarConfig.parameters.Where(param => param.input != null).ToArray<IParameterDefinition>();
 
-        foreach (var parameter in UnifiedTracking.AllParameters_v2.Concat(UnifiedTracking.AllParameters_v1).Concat(UnifiedTracking.HeadParameters).ToArray())
+        foreach (var parameter in UnifiedTracking.AllParameters)
         {
             paramList.AddRange(parameter.ResetParam(parameters));
         }
--- a/VRCFaceTracking.Core/OscQueryConfigParser.cs
+++ b/VRCFaceTracking.Core/OscQueryConfigParser.cs
@@ -52,8 +52,7 @@ public class OscQueryConfigParser
 
             // Reset all parameters
             var paramList = new List<Parameter>();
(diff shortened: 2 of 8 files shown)


VRCFaceTracking.Core/AvatarConfigFileParser.cs
```diff
@@ -68,7 +68,7 @@ public AvatarConfigParser(ILogger<AvatarConfigParser> parserLogger)
         ParameterSenderService.Clear();
         var parameters = avatarConfig.parameters.Where(param => param.input != null).ToArray<IParameterDefinition>();
 
-        foreach (var parameter in UnifiedTracking.AllParameters_v2.Concat(UnifiedTracking.AllParameters_v1).ToArray())
+        foreach (var parameter in UnifiedTracking.AllParameters_v2.Concat(UnifiedTracking.AllParameters_v1).Concat(UnifiedTracking.HeadParameters).ToArray())
```
regzo2, line 71: Looks good, we probably should make this cleaner in the future  to support more addendums to the UE parameters

VRCFaceTracking.Core/OscQueryConfigParser.cs
```diff
@@ -52,7 +52,7 @@ MulticastDnsService multicastDnsService
 
             // Reset all parameters
             var paramList = new List<Parameter>();
-            foreach (var parameter in UnifiedTracking.AllParameters_v2.Concat(UnifiedTracking.AllParameters_v1)
+            foreach (var parameter in UnifiedTracking.AllParameters_v2.Concat(UnifiedTracking.AllParameters_v1).Concat(UnifiedTracking.HeadParameters)
```
regzo2, line 55: Same as prev, review.

VRCFaceTracking.Core/Params/Data/UnifiedData.cs
```diff
@@ -83,6 +83,22 @@ public struct UnifiedExpressionShape
         public float Weight;
     }
 
+    public struct UnifiedHeadData
+    {
+        /// <summary>
+        /// Contains head rotation and position data.
+        /// Head rotation values should be normalized to a [-1, 1] range representing -90d to 90d rotation
+        /// Head position values should be normalized / capped to a [-1, 1] range and represent deviation from user origin point
```
regzo2, line 91: Not entirely sure if necessary but should we have a defined unit system for head position (eg. real world meters) for the position (I am assuming this is how most desktop apps send it though)?

VRCFaceTracking.Core/Services/ParameterSenderService.cs
```diff
@@ -24,7 +24,7 @@ public bool AllParametersRelevant
             if (AllParametersRelevantStatic == value) return;
             AllParametersRelevantStatic = value;
             SendQueue.Clear();
-            foreach (var parameter in UnifiedTracking.AllParameters_v2.Concat(UnifiedTracking.AllParameters_v1).ToArray())
+            foreach (var parameter in UnifiedTracking.AllParameters_v2.Concat(UnifiedTracking.AllParameters_v1).Concat(UnifiedTracking.HeadParameters).ToArray())
```
regzo2, line 27: Same as prev. review.

regzo2 (approved):
Looks good! Just some questions about nitpicky stuff / future changes.

Also including the head parameters in the new Data Filter mutation might be something to consider.

kusomaigo:
> Also including the head parameters in the new Data Filter mutation might be something to consider.

will do

VRCFaceTracking.Core/Params/Data/UnifiedData.cs
kusomaigo: ifacialmocap gives you a sort of nebulous value, you're right that we may want this range to be normalized to some actual movement range though, so that different modules don't result in drastically different results

PR #267, opened 2025-04-07
Improve the application shutdown logic

matty:
Previously had issues with the application not shutting down correctly, this adds logic to improve that and also adds progress of the application closing instead of "hanging".


matty:
If accepted should probably squash the commits as I forked from https://github.com/hyblocker/VRCFaceTracking and forgot so I've just updated the master from this into the branch to verify no conflicts.

PR #301, opened 2025-09-04
Fixes for network-related crashes (IPv6, VPN, multiple addresses)

AlexisEvo:
This PR tries to fix three common crashes I've seen users encounter on Discord.

1. [C#'s HttpClient does not fall back to IPv4](https://github.com/dotnet/runtime/issues/26177) if it fails to establish an IPv6 connection ("happy eyeballs"). On a poorly configured network, where the user's router provides their PC an IPv6 address, and their DNS provides an AAAA record, but they are unable to connect to an IPv6 host, this causes HttpClient to stall for 100 seconds before raising an exception. This is, unfortunately, very common. ~~An ideal solution would be to attempt to connect to IPv6, then fallback to IPv4. This PR just forces IPv4.~~ This PR now includes proper IPv6 -> IPv4 failover.
2. VRCFT often crashes when trying to setup multicast on VPN or VMware/Hyper-V/etc network adapters. Additional filtering in GetIpv4NetInterfaces to attempt to resolve this.
3. VRCFT will crash if a network interface has two IP addresses assigned to it. We only need to setup multicast on one of them.

### DIFF
--- a/VRCFaceTracking.Core/Helpers/HappyEyeballs.cs
+++ b/VRCFaceTracking.Core/Helpers/HappyEyeballs.cs
@@ -1,0 +1,254 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Net;
+using System.Net.Sockets;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace VRCFaceTracking.Core.Helpers;
+
+// Credit: https://slugcat.systems/post/24-06-16-ipv6-is-hard-happy-eyeballs-dotnet-httpclient/
+public static class HappyEyeballsHttp
+{
+    private const int ConnectionAttemptDelay = 250;
+
+#if DEBUG
+    private const int SlowIpv6 = 0;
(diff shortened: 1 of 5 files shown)


InconsolableCellist:
For #1, what happens if a user has a properly configured IPv6-only network or a network interface that doesn't have an IPv4 address assigned, just an IPv6 one? If it forces IPv4 won't it break for those users?

InconsolableCellist (approved):
The code looks good to me, and I too hate the name from the RFC, "Happy Eyeballs." Leave it to Apple.

But this does look like it'll resolve the issue of IPv6 being assigned but not usable on a network. I haven't personally had problems with VRCFT crashing due to VPNs, at least not that I've noticed. I think, however, this is ready for testing in a future pre-release.

AlexisEvo:
I sent this build to [someone on Discord that had this broken IPv6 configuration](https://discord.com/channels/849300336128032789/1414482226133074012) and they confirmed the IPv4 failover works. I tried my best to replicate it when testing this PR, but it's good to see it actually work in the wild.

Issue #305, opened 2025-10-07
[FeatureRequest] Implement mapping params to multiple addresses, and/or, always send an official param in addition to whatever is found


InconsolableCellist:
Currently BaseParam grabs the `FirstOrDefault` param that matches: https://github.com/benaclejames/VRCFaceTracking/blob/master/VRCFaceTracking.Core/OSC/DataTypes/BaseParameter.cs#L97

[In our docs](https://docs.vrcft.io/docs/tutorial-avatars/tutorial-avatars-extras/parameters) we specify that an avatar creator can use any prefix; the most popular is likely `FT`, used in Jerry's templates, but others that creators have chosen are `EchoFT`, `KIT`, or just the raw parameter names (legacy), e.g., `JawOpen`.

This means that for someone creating a Unity project that leverages face-tracking to drive behavior based on the received param names, they have to either choose the estimated most popular option, the one they personally use on their avatars, or create Unity logic to handle the options they're aware of (which is likely to be a non-exhaustive list and can complicated animator logic).

Thus it'd be nice if we: 

1. Modify the code to create a mapping not just of the first param that matches, but multiple
2. And/or always send `FT` prefixed params as well as the first one that's found that matches

This would remove the burden from asset creators and maintain backwards compatibility.

PR #306, opened 2025-10-08
305 - Sends FT prefix params in all cases where a param is sent

InconsolableCellist:
Updates the BaseParam reset logic to now provide a mapping of params to OSC addresses that meet the following criteria:

1. If a param matches (using the same previous rules), e.g., JawOpen, that param will be mapped to the OSC address as before (/avatar/parameters/JawOpen)

2. Param scanning will continue, and include all params that match--OSC messages will be sent for each of these, whereas previously the mapping would stop after finding the first match

3. If at the end of scanning a parameter is mapped but isn't destined to be sent to the default FT/ prefix (e.g., /avatar/parameters/FT/v2/JawOpen), the FT address will be included as an additional destination

4. If no parameters match, it won't also be sent to the FT address (prior behavior--doesn't send params that aren't actually used by the avatar)

This effectively makes the FT prefix the standard. See #305 for the feature request, but this is to allow avatar creators to have a reliable method to receive face-trakcing params.

### DIFF
--- a/VRCFaceTracking.Core/OSC/DataTypes/BaseParameter.cs
+++ b/VRCFaceTracking.Core/OSC/DataTypes/BaseParameter.cs
@@ -57,7 +57,7 @@ public class BaseParam<T> : Parameter where T : struct
 
     public T ParamValue
     {
-        get => (T)OscMessage.Value;
+        get => (T)OscMessages[0].Value;
         set
         {
             if (value.Equals(_lastValue))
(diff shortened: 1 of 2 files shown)


InconsolableCellist:
Tests completed: 

* Protokol, confirmed expected params being sent
* Legacy avatar, with params like `JawOpen`. Confirmed FT works
* Avatar templated with ADJerry91's template, e.g. `FT/v2/JawOpen`. Confirmed FT works
* Avatar templated with another prefix, e.g `Kit/v2/JawOpen`. Confirmed FT works
* Avatar that uses both `Kit/V2/JawOpen` and `FT/v2/JawOpen`, testing that the new param is sent in this case (the feature request). Confirmed FT works for both of those params at the same time (both sent)

benaclejames (approved):

PR #326, opened 2025-12-04
Revert calibration model

regzo2:
This is a reversion of the calibration model with some tweaks and retaining of existing features from the new calibration system. Most notable changes:

* Partial reversion back to the old calibration system
  * Uses the new calibration saturation system (parameter changes are recorded into a dataset instead of having a time limit from the old system).
  * Mostly reverted to old calibration model. 
    * Removed 'floor' calculation (this would always be 0 regardless of the interface).
  * Most users do not use calibration so it is disabled by default (interfaces should already normalized, so let users opt-in).
* Overhauled options for calibrator
  * 'Calibration Blend' slider lerps between the raw data and calibrated data. A value of 1 should be identical to the old calibrator
  * 'Continuous Calibration' toggle: Switch between rolling window (continuous) or fixed window (non-continuous) for parameter's dataset.
  
 Still need to test further to make sure it resembles the old calibrator ranges that users were happy with. The behavior of the calibrator is just normalization now. I would like to work on a 'non-linear' expression system which I think was being conflated with the calibrator but that will be separate mutator from the calibrator.

### DIFF
--- a/VRCFaceTracking.Core/OSC/DataTypes/BinaryBaseParameter.cs
+++ b/VRCFaceTracking.Core/OSC/DataTypes/BinaryBaseParameter.cs
@@ -1,4 +1,5 @@
-﻿using System.Text.RegularExpressions;
+﻿using System.Linq;
+using System.Text.RegularExpressions;
 using VRCFaceTracking.Core.Contracts;
 using VRCFaceTracking.Core.Params;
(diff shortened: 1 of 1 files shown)


InconsolableCellist (approved):
If this is just a revert to the previous logic, should be good

Issue #358, opened 2026-06-08
Question: Recommended way to post-process Unified Expressions and output custom OSC parameters


0suu:
## Question: Recommended way to post-process Unified Expressions and output custom OSC parameters

Hi, thank you for maintaining VRCFaceTracking.

I am considering a post-processing feature that reads VRCFT Unified Expressions and outputs simplified custom OSC avatar parameters.

Example flow:

```text
Tracking module, e.g. PICO / Quest Pro / Vive
  -> VRCFT Unified Expressions
  -> post-processing / classification
  -> /avatar/parameters/FER_EmotionIndex
  -> /avatar/parameters/FER_EmotionPower
```

Example output:

```text
FER_EmotionIndex : int
FER_EmotionPower : float
```

The index would represent high-level avatar states such as:

```text
0 = Neutral
1 = Joy
2 = Angry
3 = Sad
4 = Surprise
5 = Shy
6 = Smug
7 = Sleepy
```

From the documentation, I understand that the current external Tracking Module SDK is mainly intended for input modules that send tracking data into `UnifiedTracking`.

My questions are:

1. Is reading `UnifiedTracking.Data` from another external module for post-processing considered a supported/stable use case?
2. Is there a recommended way to output additional custom OSC avatar parameters from VRCFT or a module?
3. If not, would the recommended approach be an external OSC proxy/bridge instead?
4. Is a future processor/output module API something that would fit VRCFT's architecture?

The goal is to let avatar creators use a small number of stable parameters instead of implementing complex face-expression logic inside every avatar controller.

If this kind of feature would be better implemented in VRCFaceTracking Core rather than as an external module, I would also appreciate any guidance on what kind of architecture would be acceptable for a future PR.

Thank you.

benaclejames:
Hi! Thanks for your interest in contributing!

To answer your questions:
1. As far as I recall, we actually don't sync the state of our tracking data to our sandboxed tracking modules. We only sync from them, so this approach wouldn't work unless we started syncing bidirectionally.
2. We do not support outputting OSC directly from a module. It may be technically possible, but we have historically decided against such decisions as we believe it may create fractures in the UE standard and lead users to be required to use multiple different modules to use certain avatars. For an example of how this "extend the standard" problem might become an issue, see Meta's OpenXR extensions. We would, however, be open to implementing additional helper OSC messages such as the one you've mentioned to the base VRCFT application.
[...]

PR #359, opened 2026-06-15
Add optional Face Expression OSC helper output

0suu:
https://github.com/user-attachments/assets/1e64b452-f259-46e1-86e5-28381f9a2127

## Summary

This PR adds an optional Core-side OSC helper that classifies calibrated facial expression patterns from Unified Expressions and exposes the result as avatar parameters.

It is intended as an opt-in avatar helper output built from the user's calibrated facial expression templates.

## Why calibration is required

This helper uses calibration instead of a fixed rule-based mapping because face tracking behavior and intended avatar expressions can vary by user and avatar. Calibration lets the user define which facial pattern should map to each helper expression, instead of assuming one universal expression mapping.

## What changed

- Added a `Face Expression OSC Output` postprocessor mutation.
- Added helper OSC parameters:
  - `/avatar/parameters/v2/FaceExpressionIndex`
  - `/avatar/parameters/v2/FaceExpressionPower`
- Added calibration for:
  - Neutral
  - Joy
  - Angry
  - Sad
  - Surprise
- Added per-expression activation thresholds for Joy, Angry, Sad, and Surprise.
- Added saved calibration data, status rows, calibration buttons, and reset controls in the mutation UI.
- Added template-based scoring from a reduced Unified Expressions feature vector.
- Added hold/margin switching rules to reduce rapid expression flicker.
- Kept the helper disabled until Neutral, Joy, Angry, Sad, and Surprise are all calibrated.

## Behavior

Expression index values are:

- `0` = Neutral
- `1` = Joy
- `2` = Angry
- `3` = Sad
- `4` = Surprise

## Supporting changes

This PR also includes mutation UI infrastructure changes needed by the new calibration controls.

Mutation UI components are now created and refreshed through the UI dispatcher so stateful mutation controls, such as calibration status and button enabled/text updates, are marshalled back to the UI thread.

Mutation loading is now awaitable during startup, and each mutation is initialized before it is added to the active mutation collection. This avoids OSC updates observing partially loaded or uninitialized mutations during startup.

## Notes

The helper parameters currently use `v2/FaceExpressionIndex` and `v2/FaceExpressionPower`. I would appreciate feedback on whether these names fit the existing v2 parameter naming scheme, or if a different naming convention would be preferred.

This PR extends the mutation UI support layer because the calibration workflow needs to display current calibration state, button text, and enabled/disabled state dynamically. I kept the new components generic (`MutationInfo`, `MutationStatus`, `MutationStatusAction`) instead of making them face-expression-specific, and would appreciate feedback on whether this is the right level of abstraction for the project.

Related to #358.

### DIFF
--- a/VRCFaceTracking.Core/Contracts/Services/IDispatcherService.cs
+++ b/VRCFaceTracking.Core/Contracts/Services/IDispatcherService.cs
@@ -5,4 +5,9 @@
 public interface IDispatcherService
 {
     public void Run(Action action);
+
+    /// <summary>
+    /// Runs an action on the UI dispatcher and completes after the action has run.
+    /// </summary>
+    public Task RunAsync(Action action);
 }
--- a/VRCFaceTracking.Core/MainStandalone.cs
+++ b/VRCFaceTracking.Core/MainStandalone.cs
@@ -49,7 +49,7 @@ public class MainStandalone : IMainService
         _logger.LogDebug("Teardown complete. Awaiting exit...");
     }
 
-    public Task InitializeAsync()
+    public async Task InitializeAsync()
     {
         VRChat.EnsureVRCOSCDirectory();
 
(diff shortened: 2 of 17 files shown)


0suu:
@benaclejames 

Hi, just checking in on this PR.

I realize it is a fairly large PR. If it would be easier to review, I can split it into smaller PRs, for example the mutation UI / initialization changes first, and the Face Expression OSC helper separately.

I would appreciate any feedback on whether the overall direction fits the project. Thanks!

PR #372, opened 2026-09-05
Avalonia UI switch

benaclejames:

PR #374, opened 2026-09-06
Add module state switching to the Module Registry

moewhite19:
Adds a per-module state selector to the Module Registry so tracking modules can be mixed or switched without uninstalling.

Each installed module can be set to one of:

    Enabled
    Disabled
    Eyes Only
    Face Only

Disabled modules are skipped on load; Eyes Only / Face Only restrict which tracking slot a module may claim. The module list shows the current state and marks pending changes until the app is restarted.


benaclejames:
Generally we do not accept PRs created entirely by Artificial Intelligence. Can you confirm if this is the case with this PR as it has a couple of areas that look as if this may be the case.

Additionally, this PR targets the master branch, but I worry that the moment we merge Avalonia, it'll be broken. Apologies in advance if this is the case as we will be merging avalonia to master in the coming week. You can view the progress of said PR at #372

moewhite19:
AI was only used at the start to speed up the initial setup – everything after that was manually maintained by me, though the PR description and commits were AI-generated for efficiency.

As for Avalonia – I couldn't get it to run on Steam, so I haven't been able to test it. I think it's better to wait until Avalonia is stable and merged into master, and then I'll continue maintaining the fork accordingly.

